PR #18, opened 2021-05-17
Add Pull Request template

joker234:


matthiasschaub:
Do I need to copy and paste the WIP symbol 🚧? I find it more convinient if I just  have to write WIP or something on my keyboard instead of navigating to another PR and copy & paste the symbol.

joker234:
> Do I need to copy and paste the WIP symbol construction? I find it more convinient if I just have to write WIP or something on my keyboard instead of navigating to another PR and copy & paste the symbol.

That's personal preference and up to you. I like the :construction: symbol more, because it stands out in the list of PRs. Furthermore, this is the same text as for OSHDB an ohsome API. I would keep it like that and people can use whatever they like ;)

matthiasschaub (dismissed):

matthiasschaub (approved):

PR #28, opened 2021-05-19
Check for invalid indicator layer combination during initialization of indicator objects

### LINKED ISSUE #24: Find and fix test causing a 400 error from ohsome API
One test in `test_indicator_tags_ratio.py` seem to cast a request to the ohsome API with the parameter `filter2=None` which is not a [valid input](https://docs.ohsome.org/ohsome-api/stable/filter.html).

See the request here: [test_indicator_tags_ratio.json](https://github.com/GIScience/ohsome-quality-analyst/blob/main/workers/tests/integrationtests/fixtures/vcr_cassettes/test_indicator_tags_ratio.json#L270-L315)

See also #23

matthiasschaub:
### Corresponding issue
Closes #24 

### Checklist
- [x] Squash commits

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Current Main
 
+- Add check and custom exception for an invalid indicator layer combination during initialization of indicator objects ([#28])
+
+[#28]: https://github.com/GIScience/ohsome-quality-analyst/pull/28
+
 
 ## 0.3.1
 
--- a/workers/tests/integrationtests/test_indicator_tags_ratio.py
+++ b/workers/tests/integrationtests/test_indicator_tags_ratio.py
@@ -38,6 +38,7 @@ class TestIndicatorRatio(unittest.TestCase):
         """Ratio should be 1.0 when all features match expected tags"""
         layer_name = "jrc_health_count"
         dataset = "regions"
+        # TODO: Choose smaller region
         feature_id = 9  # DE Aschaffenburg (good)
         bpolys = asyncio.run(db_client.get_bpolys_from_db(dataset, feature_id))
 
@@ -67,13 +68,11 @@ class TestIndicatorRatio(unittest.TestCase):
         """Layer with no filter2 for ratio endpoint"""
         layer_name = "major_roads_length"
         dataset = "regions"
-        feature_id = 2
+        feature_id = 28
         bpolys = asyncio.run(db_client.get_bpolys_from_db(dataset, feature_id))
 
-        indicator = TagsRatio(layer_name=layer_name, bpolys=bpolys)
-        asyncio.run(indicator.preprocess())
-        self.assertIsNone(indicator.count_all)
-        indicator.calculate()
+        with self.assertRaises(ValueError):
+            TagsRatio(layer_name=layer_name, bpolys=bpolys)
 
 
 if __name__ == "__main__":
--- a/workers/ohsome_quality_analyst/base/indicator.py
+++ b/workers/ohsome_quality_analyst/base/indicator.py
@@ -13,7 +13,11 @@ import matplotlib.pyplot as plt
 from dacite import from_dict
 from geojson import FeatureCollection
 
-from ohsome_quality_analyst.utils.definitions import get_layer_definition, get_metadata
+from ohsome_quality_analyst.utils.definitions import (
+    INDICATOR_LAYER,
+    get_layer_definition,
+    get_metadata,
+)
 
 
 @dataclass
@@ -81,6 +85,8 @@ class BaseIndicator(metaclass=ABCMeta):
             svg=self._get_default_figure(),
         )
 
+        self._validate_indicator_layer(self.__class__.__name__, layer_name)
+
     @abstractmethod
     async def preprocess(self) -> bool:
         """Get fetch and prepocess data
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Current Main
 
+- Ceck for invalid indicator layer combination during initialization of indicator objects ([28])
```
joker234, line 5:
```suggestion
- Add check for an invalid indicator layer combination during initialization of indicator objects ([#28])
```

workers/tests/unittests/test_custom_exceptions.py
```diff
@@ -0,0 +1,11 @@
...
+
+from ohsome_quality_analyst.utils.exceptions import IndicatorLayerError
+
+
+class TestCustomException(unittest.TestCase):
+    def test(self):
+        with self.assertRaises(IndicatorLayerError):
+            raise IndicatorLayerError(("Indicator", "Layer"))
+        with self.assertRaises(ValueError):
+            raise IndicatorLayerError(("Indicator", "Layer"))
```
joker234, line 11: I'm not 100% sure what you want to test here by using `("Indicator", "Layer")` but if you intend to give the Exception/Error more context, I would
1. add a description to the `IndicatorLayerError` definition, and
2. check these arguments in the tests as well, e.g. like this:
```suggestion
        indicator = "MyIndicator"
        layer = "MyLayer"
        with self.assertRaises(IndicatorLayerError) as cm:
            raise IndicatorLayerError((indicator, layer))
        self.assertEqual(cm.exception.args[0], (indicator, layer))
        with self.assertRaises(ValueError) as cm:
            raise IndicatorLayerError((indicator, layer))
        self.assertEqual(cm.exception.args[0], (indicator, layer))
```

workers/tests/integrationtests/test_indicator_tags_ratio.py
```diff
@@ -62,19 +62,6 @@ def test_no_features(self):
...
-        layer_name = "major_roads_length"
-        dataset = "regions"
-        feature_id = 2
-        bpolys = asyncio.run(db_client.get_bpolys_from_db(dataset, feature_id))
-
-        indicator = TagsRatio(layer_name=layer_name, bpolys=bpolys)
-        asyncio.run(indicator.preprocess())
-        self.assertIsNone(indicator.count_all)
-        indicator.calculate()
-
```
joker234, line 77: Which test replaces this test? Is there another test checking if a ratio query is made without a `filter2` or `ratio_filter` argument?

Please remove the referring query in the cassettes file as well.

workers/ohsome_quality_analyst/utils/exceptions.py
```diff
@@ -0,0 +1,7 @@
+"""Custom exception classes"""
+
+
+class IndicatorLayerError(ValueError):
+    """Indicator layer combination is invalid"""
```
joker234, line 5: See comment made to the newly introduced tests.

joker234 (changes requested):
Some minor suggestions. Thx for implementing this check.

In general it would be nice if the (API) user would get more feedback about the actual exceptions. Most of the time they get `Internal Server Error`. See also https://github.com/GIScience/ohsome-quality-analyst/pull/29#discussion_r636138108

workers/tests/integrationtests/test_indicator_tags_ratio.py
matthiasschaub: The Tag Ratio indicator works only for layer with the `ratio_filter`. Since changes introduced in this PR indicator + layer combination are validated during initialization of the indicator object. This will cover up layer with not  `ratio_filter` for the Tag Ratio indicator.

For this particular test case a `IndicatorLayerError` will be raised before the test finishes.

workers/tests/unittests/test_custom_exceptions.py
matthiasschaub: I added your suggestion to the tests.

Also I improved the exception message with a description + indicator layer tuple of validation function in base/indicator.py

workers/tests/integrationtests/test_indicator_tags_ratio.py
matthiasschaub: I recreated the caassette for this test.

workers/tests/unittests/test_custom_exceptions.py
joker234: Okay, as far as I understand your code now, `IndicatorLayerError` should only take a string and not a tuple. I think you should change the tests then as well to check if the string matches (or get rid of the check again, as you only test the behaviour of `ValueError`).

Something like this:
```python
        with self.assertRaises(IndicatorLayerError) as cm:
            raise IndicatorLayerError("foo")
        self.assertEqual(cm.exception.args[0], "foo")

matthiasschaub: Tuple would be fine. It would be cast to the string representation by the exception class. But you are right in that it is not needed here. I changed it to use a string as you suggested it

joker234: Hmm okay, I didn't mean to get rid of the whole `IndicatorLayerError`, but okay ;)

joker234 (dismissed):

matthiasschaub:
@joker234 since I rebased (squashed) the commits your approval is not longer sufficient for GitHub to let me merge the PR. I did not make any changes to the code. Could you again approve the PR?

joker234 (approved):

PR #48, opened 2021-06-09
Fix spacing in exception message

matthiasschaub:


Hagellach37 (approved):
Not sure how strict we have to be with reviews, but for me something like this could be merged directly.

PR #52, opened 2021-06-09
🚧 Fixing Bug in Report RemoteMappingLevelOne creation

hn437:
### Description
The creation of the Report RemoteMappingLevelOne fails in regions where no mapping has taken place.
Report RemoteMappingLevelOne should be set to the label undefined if all indicators are undefined

### Corresponding issue
Closes #32 

### Checklist
- [x] I have updated my branch to `main` (e.g. through `git rebase main`)
- [x] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#pre-commit) before committing
- [ ] I have commented my code

PR #53, opened 2021-06-10
Implement combine_indicators() as Concrete Method

### LINKED ISSUE #32: Creation of a report fails when no mapping has happened in the region
Creating a report like remote_mapping_level_one calculates the mean value when combining the indicators. However, if no mapping has taken place, the indicators are labeled undefined, and no undefined indicators are considered. Therefore, the mean calculation fails, as it needs at least one input. This error aborts the calculation for all further report creations as well

matthiasschaub:
### Description

Implement combine_indicators() as Concrete Method of the Base Blass `Report`.
This provides a default method for combining indicators to all reports.
This will also reduce code duplication across those report (child classes).

Closes #32 

### Todo

- [x] Include fix of issue #32 by implementing solution of PR #52 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add option to select different datasets and fid fields as input to OQT ([#4])
 - Rewrite save and load indicator from database logic to use one result table for all indicator results ([#37])
 - Remove GUF Comparison indicator ([#55])
+- Implement combine_indicators() as Concrete Method of the Base Blass Report ([53])
 
 [#28]: https://github.com/GIScience/ohsome-quality-analyst/pull/28
 [#40]: https://github.com/GIScience/ohsome-quality-analyst/pull/40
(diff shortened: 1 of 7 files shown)


Hagellach37 (commented):
Hey,
In this case here is there any particular reason why you still keep the `combine_indicators` function on the individual report level. So didn't you also delete the function in e.g. map_action_poc/report.py but use `super().combine_indicators()`?

Maybe you can briefly explain your thoughts and might add a comment to the code.

matthiasschaub:
>  In this case here is there any particular reason why you still keep the `combine_indicators` function on the individual report level. So why didn't you also delete the function in e.g. map_action_poc/report.py but use `super().combine_indicators()`?

For now we could also delete the function in each of the subclasses/reports and it would be the same outcome.
My thoughts were that the combine function of a report could have a different implementation then another report (E.g. using median instead of mean).
For this use case a Concrete Method is ideal. It provides a default implementation of the function while allowing overriding the function with a custom implementation in the child class.

But since this is not a part on which we are working right now, should I further simplify the code and substitute the Concrete Method implementation with a 'normal' function?

Hagellach37:
I think it's okay as it is, and maybe good to have this option in the future.

Hagellach37 (approved):

PR #57, opened 2021-06-16
Writing GeoJSON from the CLI for indicator/report calculation when using datasets

### LINKED ISSUE #10: CLI: `oqt create-indicator` should always output a GeoJSON
**Currently** the output of `oqt create-indicator --infile ...` is a GeoJSON.

**Expected** as output of `oqt create-indicator --dataset ...` is also a GeoJSON. This is not the case.

> The same is true for `oqt create-report`

hn437:
### Description
added the functionality of writing a geojson as output for calculatg indicators/reports when the dataset is used instead of an infile. Only applicable if the outfile flag is set.

### Corresponding issue
Closes #10

### Checklist


workers/ohsome_quality_analyst/cli.py
```diff
@@ -227,9 +239,23 @@ def create_report(
...
+                feature["properties"].update(vars(report.metadata))
+                feature["properties"].update(vars(report.result))
         click.echo(report.metadata)
         click.echo(report.result)
+    if outfile:
+        """Writing the output GeoJSON"""
+        outfile = pathlib.Path(outfile)
+        outfile.parent.mkdir(parents=True, exist_ok=True)
+        with open(outfile, "w") as f:
+            geojson.dump(feature_collection, f, default=datetime_to_isostring_timestamp)
```
hn437, line 258: This is done for indicators and reports. should it be moved to the [utils/helper](https://github.com/GIScience/ohsome-quality-analyst/blob/main/workers/ohsome_quality_analyst/utils/helper.py) script as a function and called at both instances?

workers/ohsome_quality_analyst/cli.py
```diff
@@ -147,12 +149,8 @@ def create_indicator(
             feature["properties"].update(vars(indicator.result))
         if outfile is None:
             outfile = infile.stem + "_" + indicator_name + infile.suffix
```
Hagellach37, line 151: I'm not sure but this looks a bit hacky to me. Maybe add a comment here in the code how the outfile name will look like based on what is provided as input file.

workers/ohsome_quality_analyst/cli.py
```diff
@@ -147,12 +149,8 @@ def create_indicator(
...
-        else:
-            outfile = pathlib.Path(outfile)
-            outfile.parent.mkdir(parents=True, exist_ok=True)
-        with open(outfile, "w") as f:
-            geojson.dump(feature_collection, f, default=datetime_to_isostring_timestamp)
     else:
+        """When using a dataset and FID as input"""
```
Hagellach37, line 153: see above about `#`

workers/ohsome_quality_analyst/cli.py
```diff
@@ -165,9 +163,25 @@ def create_indicator(
...
+            )
+            for feature in feature_collection.features:
+                if indicator.data is not None:
+                    feature["properties"].update(vars(indicator.data))
+                feature["properties"].update(vars(indicator.metadata))
+                feature["properties"].update(vars(indicator.result))
         click.echo(indicator.metadata)
         click.echo(indicator.result)
+    if outfile:
+        """Writing the output GeoJSON"""
```
Hagellach37, line 180: I think it is good to put this into a function in utils.

workers/ohsome_quality_analyst/cli.py
```diff
@@ -165,9 +163,25 @@ def create_indicator(
                 force=force,
             )
         )
-        # TODO: Save as GeoJSON instead of printing to Stdout
+        if outfile:
```
Hagellach37, line 166: Hey,
to be honest I'm a bit confused by the many if else statements in this function. Let's think about a way to make this eaier to read. Could you try to explain again what the major things are that you try to cover here and maybe this will help us to reduce the complexity of the code.

For instance here you check check `if outfile:` and a few lines later in line 179 you check it again.

workers/ohsome_quality_analyst/cli.py
Hagellach37: yep putting into a helper function would be nice.

Hagellach37:
Hey @hn437 ,
thanks for taking a look into this. :) In general I think that it does what we want to achieve, but at the current state it looks to me rather complex and maybe we can find a way to implement it so that it is easier to understand.

So basically we want to distinguish the following scenarios:
* user provides input file versus user passed FID and dataset name
* user wants to print to stdout or user wants to also save as geojson file

I would suggest that we also split this better in out code and try not to merge these parts. So in the first part only take of the calculation of the indicator. And then only later check if the results should be saved into a geojson file and only if so generate the output filename etc.

How does that sound to you?

workers/ohsome_quality_analyst/cli.py
```diff
@@ -123,6 +124,7 @@ def create_indicator(
         )
         click.confirm("Do you want to continue?", abort=True)
     if infile is not None:
+        """When using an infile as input"""
         infile = pathlib.Path(infile)
         with open(infile, "r") as file:
```
Hagellach37, line 129: maybe also this part could be moved into a function called `load_infile` or similar. since it is used here and in the report as well. And in this function you can also do the check if the geometry is valid.

workers/ohsome_quality_analyst/cli.py
hn437: okay, will do that!

workers/ohsome_quality_analyst/cli.py
```diff
@@ -123,6 +124,7 @@ def create_indicator(
         )
         click.confirm("Do you want to continue?", abort=True)
     if infile is not None:
+        """When using an infile as input"""
```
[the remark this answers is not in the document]
hn437: will change it!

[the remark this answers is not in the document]
matthiasschaub: Thats right. The notation using `"""` is called docstring. See also: https://www.python.org/dev/peps/pep-0257/
> A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition. Such a docstring becomes the __doc__ special attribute of that object.

For comments just use `#`

workers/ohsome_quality_analyst/cli.py
hn437: this is included for a while now, not with this PR. It was the best way @matthiasschaub and I found to automatically create an outputfile definition which works both on linux and windows, due to the different handling of subdirs. But I'll add a comment what is done here... alternatively, I could also change it that an output geojson will only be written if the outfile is defined by the user

workers/ohsome_quality_analyst/cli.py
hn437: will change it as well

workers/ohsome_quality_analyst/cli.py
hn437: will do it!

workers/ohsome_quality_analyst/cli.py
hn437: I will add this function and replace the code

workers/ohsome_quality_analyst/cli.py
Hagellach37: I think it's fine if you add a comment stating just what you explained to me here. That you do it like this with good reasons.

workers/ohsome_quality_analyst/cli.py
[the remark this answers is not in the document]
hn437: I will keep it in mind!

workers/ohsome_quality_analyst/cli.py
hn437: is the comment I added understandable?

Hagellach37: yes

workers/ohsome_quality_analyst/cli.py
```diff
@@ -141,18 +143,14 @@ def create_indicator(
...
             )
-            if indicator.data is not None:
-                feature["properties"].update(vars(indicator.data))
-            feature["properties"].update(vars(indicator.metadata))
-            feature["properties"].update(vars(indicator.result))
+            feature = update_features_indicator(feature, indicator)
         if outfile is None:
+            # the outfile will be written in the same dir as the infile. the original
+            # name will be kept, but the name of the indicator will be added to it
             outfile = infile.stem + "_" + indicator_name + infile.suffix
```
Hagellach37, line 152: Hey @hn437 @matthiasschaub ,
this is not working for me as expected. When I run the cli the output file will be written to my current working directory, but the location of the input file.

```
:~/GIT_REPOS/ohsome-quality-analyst/workers$ oqt create-indicator -i GhsPopComparisonBuildings -l building_count --infile=/home/h9b/Downloads/test_regions_oqt.geojson
(ohsome-quality-analyst-eBugwxcL-py3.8) h9b@h9b-ThinkPad-T480s:~/GIT_REPOS/ohsome-quality-analyst/workers$ 

ls
Dockerfile  Dockerfile.continuous-integration  ohsome_quality_analyst  poetry.lock  pyproject.toml  scripts  setup.cfg  test_regions_oqt_GhsPopComparisonBuildings.geojson  tests

```

Hagellach37:
Hey,
in general the changes look good to me now. I've just added a minor change to fix an error when handling the infile string.

@matthiasschaub @hn437 I'm not sure if you already planned, but I think we also need to add a test to `test_cli.py`. Currently this tests only for features that are in the database.

On another note I'm wondering if we want to export the fields `label_description` and `result_description` in the geojson file. These are not really useful to my opinion since they just contain the placeholders strings and not the actual values. Also it's somehow duplicated to what we have in the field `description`.

Not sure if we want to address it in this merge request.

image

hn437:
> Hey,
> in general the changes look good to me now. I've just added a minor change to fix an error when handling the infile string.
> 
> @matthiasschaub @hn437 I'm not sure if you already planned, but I think we also need to add a test to `test_cli.py`. Currently this tests only for features that are in the database.
> 
> On another note I'm wondering if we want to export the fields `label_description` and `result_description` in the geojson file. These are not really useful to my opinion since they just contain the placeholders strings and not the actual values. Also it's somehow duplicated to what we have in the field `description`.
> 
> Not sure if we want to address it in this merge request.
> 
> image

yes, thank you! That change of type from string to part was exported into the helper function but is needed there, forgot to implement it again.

a test is planned, also as the function to calculate for files with multiple features needs to be tested as well. but this will be another PR.

We should also look into what information is added to the features, yes. at the moment all entries in the dicts result and metadata are added, but I agree, maybe not all entries are needed. we also wanted to add the specific values (like saturation or growth value) to the data dict, maybe we could do this all at once in another PR

workers/ohsome_quality_analyst/cli.py
hn437: it worked for me before I used the helper functions. maybe something was corrupted by using them, I'll look into it on monday. thank you for letting me know!

hn437: I adressed it in the latest commit. Now, it works for me on Windows, maybe you could check it for linux @Hagellach37 and let me know if it works there as well? Thank you!

matthiasschaub:
@Hagellach37 @hn437 

Tests for testing file input and output to the CLI are planned. Those are probably not that straightforward. I am happy to work with @hn437 on this.

Regarding the output GeoJSON schema I think something like the following would fit our current structure (e.g. Indicator class):

```json
{
    "type": "Feature",
    "geometry": {
        "type": "MultiPolygon",
        "coordinates": [ ... ]
    },
    "properties": {
        "metadata": {
            "name": ...,
            "description": ...
        },
        "layer": {
            "name": ...,
            "description": ...,
            "endpoint": ...,
            "filter": ...,
            "ratio_filter": ...
        },
        "result": {
            "timestamp_oqt": ...,
            "timestamp_osm": ...,
            "value": ...,
            "label": ...,
            "description": ...,
            "svg": ...
        },
        "data": {
            "misc": ...
        }
    }
}
```

hn437:
> @Hagellach37 @hn437
> 
> Tests for testing file input and output to the CLI are planned. Those are probably not that straightforward. I am happy to work with @hn437 on this.
> 
> Regarding the output GeoJSON schema I think something like the following would fit our current structure (e.g. Indicator class):
> 
> ```json
> {
>     "type": "Feature",
>     "geometry": {
>         "type": "MultiPolygon",
>         "coordinates": [ ... ]
>     },
>     "properties": {
>         "metadata": {
>             "name": ...,
>             "description": ...
>         },
>         "layer": {
>             "name": ...,
>             "description": ...,
>             "endpoint": ...,
>             "filter": ...,
>             "ratio_filter": ...
>         },
>         "result": {
>             "timestamp_oqt": ...,
>             "timestamp_osm": ...,
>             "value": ...,
>             "label": ...,
>             "description": ...,
>             "svg": ...
>         },
>         "data": {
>             "misc": ...
>         }
>     }
> }
> ```

So it's probably best use to add the test within this PR, right? I would be thankful for a bit of help with them, @matthiasschaub  I'll text you so we could may coordinate when/how to do it.

I like the proposed output GeoJSON schema, and as I said on monday using the data class outside of the result class would make it easy to add it to the output and Is already covered by the helper-function. only need to add the respective data from each indicator to the data class and add it in the base class.

regarding the descriptions @Hagellach37 mentioned: the `label_description` and `result_description` are part of the metadata and show the possible results that may occur, the label and result are part of the results and show the actual class out of the possible ones for the feature. Having both or only the results is probably depending on the prefered style, should be easy to change if only the results are wanted. I think it's good to know which classes may appear and what they stand for, however, this does not necessarily mean it has to be written as an attribute to the geojson. It could also be done either on the webpage or a doc with the definitions. I think this is a point up for discussion...

and finally, I just noticed a bug (and will open an issue for this): the metadata contains an attribute `description` which describes the indicator, and the results contain an attribute `description` as well, which describes the result. Due to the matching name, the result-description overwrites the indicator-description.

workers/ohsome_quality_analyst/cli.py
```diff
@@ -114,7 +118,8 @@ def create_indicator(
     fid_field: str,
     force: bool,
 ):
-    """Create an Indicator and print results to stdout."""
+    """Create an Indicator and print results to stdout. Write a GeoJSON if an outfile
+    is specified or an infile is used as input"""
```
matthiasschaub, line 122: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. 

```suggestion
    """Create an Indicator.
    
    Write a GeoJSON with the result to disk if an outfile is specified.
    Otherwise print to stdout.
    """
```

matthiasschaub (dismissed):

matthiasschaub (dismissed):

matthiasschaub (approved):

Issue #46, opened 2021-06-07
CLI/API output/response should have the same schema


matthiasschaub:
CLI and API should both output a JSON (when input is a feature of a dataset) or GeoJSON (if input is bpolys) with a shared schemata.

matthiasschaub:
This could be archived by implementing the __geo_interface__ for the indicator and report classes:
- https://pypi.org/project/geojson/#custom-classes
- https://gist.github.com/sgillies/2217756

```python
class BaseIndicator(metaclass=ABCMeta):

    def __init__():
        ...

    @property
    def __geo_interface__(self):
        return {
            "type": "Multipolygon",
            "coordinates": self.bpolys,
            "properties": {vars(sel.result), self.layer.name, self.metadata.name},
        }
```

matthiasschaub:
Similar to above proposal the `__str__()` method could be implemented. This could be used by the CLI to pretty print a indicator or object if a GeoJSON is not needed.

matthiasschaub:
CLI response should look something like this:

```json
{
    "type": "Feature",
    "geometry": {
        "type": "MultiPolygon",
        "coordinates": [ ... ]
    },
    "properties": {
        "metadata": {
            "name": ...,
            "description": ...
        },
        "layer": {
            "name": ...,
            "description": ...,
            "endpoint": ...,
            "filter": ...,
            "ratio_filter": ...
        },
        "result": {
            "timestamp_oqt": ...,
            "timestamp_osm": ...,
            "value": ...,
            "label": ...,
            "description": ...,
            "svg": ...
        },
        "data": {
            "misc": ...
        }
    }
}
```

PR #66, opened 2021-06-24
Implement `__geo_interface__` for indicator class

matthiasschaub:
### Description

Implement `__geo_interface__` for indicator class.

This solution came up in discussion of issue #46 

### Checklist

### Blocker
- [ ] #69

### DIFF
--- a/workers/ohsome_quality_analyst/base/indicator.py
+++ b/workers/ohsome_quality_analyst/base/indicator.py
@@ -11,7 +11,7 @@ from typing import Dict, Literal, Optional
 
 import matplotlib.pyplot as plt
 from dacite import from_dict
-from geojson import FeatureCollection
+from geojson import Feature, FeatureCollection
 
 from ohsome_quality_analyst.utils.definitions import (
     INDICATOR_LAYER,
@@ -54,7 +54,6 @@ class Result:
     value: Optional[float]
     description: str
     svg: str
-    data: Optional[dict] = None
 
 
 class BaseIndicator(metaclass=ABCMeta):
@@ -64,11 +63,9 @@ class BaseIndicator(metaclass=ABCMeta):
         self,
         layer_name: str,
         bpolys: FeatureCollection = None,
-        data: dict = None,
     ) -> None:
 
         self.bpolys: str = bpolys
-        self.data: Optional[dict] = data
 
         # setattr(object, key, value) could be used instead of relying on from_dict.
         metadata = get_metadata("indicators", type(self).__name__)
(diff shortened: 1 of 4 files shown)


matthiasschaub:
@Hagellach37 @hn437, 

This is a proof of concept implementation of the `__geo_interface__` (https://github.com/jazzband/geojson) for the indicator class.
I find this solution elegant and simple. 

This would make the code in CLI/API simpler since they do not have to bother with building the GeoJSON/JSON.
Also if we decide to return a GeoJSON as response from the API this would make the responses of API and CLI uniform.

Im am curious what you two think about this approach.

matthiasschaub:
After using geojson.dumps on the indicator object the resulting GeoJSON should look like this (This is currently not the case because bpolys is of type FeatureCollection and not of type Geometry. But this is already addressed in another PR): 
```
{
  "type": "Feature",
  "geometry": {
    "type": "MultiPolygon",
    "coordinates": [
      [
        [
          [
            10.0282,
            32.202343
          ],
          [
            10.03423,
            32.202343
          ],
          [
            10.03423,
            32.210386
          ],
          [
            10.0282,
            32.210386
          ],
          [
            10.0282,
            32.202343
          ]
        ]
      ]
    ]
  },
  "properties": {
    "name": "GHS-POP Comparison Buildings",
    "description": "Comparison between population density and feature density.\nThis can give an estimate if mapping has been completed.\n",
    "layer.name": "Building Count",
    "layer.description": "All buildings as defined by all objects tagged with 'building=*'.\n",
    "result.timestamp_oqt": "2021-06-24T08:45:34.339165",
    "result.timestamp_osm": null,
    "result.label": "green",
    "result.value": 1,
    "result.description": "Following the GHS POP dataset, there are \n84 people living in an area of \n0.5 sqkm, which results in a population density \n165.4 of people per sqkm.\n106.5 buildings per sqkm mapped. \nFor the given population density, this is a relatively high value and \nindicates a good data quality in terms of completeness.\n"
  }
}
```

hn437:
So the big advantage would be the return of a ready to dump object frim the indicator, do I understand this correctly? So a geometry or feature as input could be directly dumped using the return?
At a first glance this looks good to me, but there are several things we would need to check if they are still easily accomplishable when using the __geo_interface__. 
However, I haven't looked into it in detail yet, so maybe these things are easy to do, just to keep them in mind:

- If the input feature contains further attributes, they may should be kept (e.g. a certain identifier), but only the geometry will be given to the indicator in future, right? so the return would need to be updateable with these original attributes
- having a feature collection with multiple features as input, only one output file shouls be returned. So you can just "collect" these returns (and add +them one after the other) in a featurecollection, right?

matthiasschaub:
Close in favor of #86

PR #37, opened 2021-05-26
Add option to select different datasets and fid fields as input to OQT

matthiasschaub:
### Description

Add option to select different datasets and fid fields as input to OQT.

Available datasets and possible fid fields are defined in `definitions.py`.

Each dataset defaults to a primary fid field. The usage of other fields as identifier is possible.
The attribute/parameter/argument `fid_field`/`fidField` is added to the report class, oqt functions and API/CLI.

Rewrite save and load indicator from database logic to use one result table for all indicator results.

### Corresponding issue
Closes #4 

### Checklist
- [x] I have updated my branch to `main` (e.g. through `git rebase main`)
- [x] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#pre-commit) before committing
- [x] I have commented my code
- [x] I have added sufficient unit and integration [tests](https://github.com/GIScience/ohsome-quality-analyst/blob/main/docs/development_setup.md#tests)
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CHANGELOG.md)
- [x] Check changes to VCR cassetts
- [x] Add sensible commits (squash, rebase, …)

PR #69, opened 2021-06-28
Bpoly feature

### LINKED ISSUE #9: The attribute `bpolys` of indicators should be of type Feature or MultiPolygon not FeatureCollection
Should an indicator object only have the geometry as attribute or the whole feature (GeoJSON Feature object) including the geometry as attribute?

Also is `bpolys` a good variable name or should it be renamed to `geometry` or `feature`?

matthiasschaub:
### Description
- Change type of attribute bpolys to be GeoJSON.Feature.
- Input bpolys to API can be of type GeoJSON.Feature with the geometry type Polygon or MultiPolygon. Input can also only contain the geometry.
- Input file to API can be any valid GeoJSON. Geometry type has to be Polygon or MultiPolygon
- Improve type hints.

### Corresponding issue
Closes #9

### Checklist
- [x] support `FeatureCollection` with only one `Feature` (again)


workers/ohsome_quality_analyst/api.py
```diff
@@ -56,26 +30,74 @@ async def load_bpolys(bpolys: str) -> None:
 class IndicatorParameters(BaseModel):
     bpolys: Optional[str] = None
     dataset: Optional[str] = None
-    featureId: Optional[Union[int, str]] = None
+    featureId: Union[int, str, None] = None
```
joker234, line 33: While this change is technically correct, the purpose has been better understood before.

The same applies to numerous occurrences in this PR.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Simplify CLI option handling by only allowing one option at a time to be added ([#54])
 - Redefine OQT regions ([#26])
 - Use ohsome API endpoint `/contributions/latest/count` for Last Edit indicator ([#68])
+- Change type of attribute bpolys to be GeoJSON.Feature ([#69] [#9])
```
joker234, line 21: This change requires a **breaking** change. Please sort this accordingly.

workers/ohsome_quality_analyst/api.py
```diff
@@ -56,26 +30,74 @@ async def load_bpolys(bpolys: str) -> None:
...
+
+def int_or_str_param_type(value: str) -> Union[int, str]:
+    """Parse parameter as Interger otherwise parse as String"""
+    try:
+        return int(value)
+    except (ValueError, TypeError):
+        if isinstance(value, str):
+            return value
+        else:
+            raise ValueError("Given parameter is not a valid integer or string")
```
joker234, line 63: Do you have an example where this could be required? As far as I see, this is only used for the `feature_id`. The feature id given by a GET/POST request should always be given as a string, afaik. If you want to prevent the parameter from being optional, I would check/set this with fastAPI.

workers/ohsome_quality_analyst/api.py
```diff
@@ -56,26 +30,74 @@ async def load_bpolys(bpolys: str) -> None:
...
+    elif isinstance(bpolys, Feature):
+        await check_geom_size(bpolys.geometry)
+        return bpolys
+    elif isinstance(bpolys, (Polygon, MultiPolygon)):
+        await check_geom_size(bpolys)
+        return Feature(geometry=bpolys)
+    else:
+        raise ValueError(
+            "Input GeoJSON Objects have to be of type Feature, Polygon or MultiPolygon"
+        )
```
joker234, line 90: Why do we disallow `FeatureCollection` here? Wouldn't it be sufficient to check if we have multiple Features?

workers/ohsome_quality_analyst/api.py
```diff
@@ -152,14 +177,21 @@ async def _fetch_report(
...
     if bpolys is not None:
         bpolys = await load_bpolys(bpolys)
 
+    if feature_id:
+        feature_id = int_or_str_param_type(feature_id)
+
     report = await oqt.create_report(
-        name, bpolys=bpolys, dataset=dataset, feature_id=feature_id, fid_field=fid_field
+        name,
+        feature=bpolys,
```
joker234, line 191: the name `feature` can be confusing, because `bpolys` might contain a `Feature` or a `Geometry`

workers/ohsome_quality_analyst/base/indicator.py
```diff
@@ -63,19 +63,19 @@ class BaseIndicator(metaclass=ABCMeta):
     def __init__(
         self,
         layer_name: str,
-        bpolys: FeatureCollection = None,
-        data: dict = None,
+        feature: Feature,
```
joker234, line 66: I don't think this is true. In the `api.py` you check if `feature` is a `Feature` or a `Geometry`.

```suggestion
        feature: Union[Feature, Geometry],
```
(don't forget the import)

workers/ohsome_quality_analyst/geodatabase/client.py
```diff
@@ -131,7 +131,7 @@ async def get_feature_ids(dataset: str, fid_field: str) -> List[Union[int, str]]
 
 
 # TODO Rewrite to work with geojson.Feature as input
-async def get_area_of_bpolys(bpolys: Dict):
+async def get_area_of_bpolys(bpolys: Union[Polygon, MultiPolygon]):
```
joker234, line 134: Do you only allow Polygons here? You could use `bpolys: Geometry` instead.

workers/ohsome_quality_analyst/api.py
matthiasschaub: In contrast to the CLI there exists no logic to handle FeatureCollection in the API. I want to address this separately after geo_interface for indicators/reports is implemented.

So for now FeatureCollection is not allowed as input using the API.

workers/ohsome_quality_analyst/api.py
matthiasschaub: Yes you are right. The else statement is never reached. Except maybe when None is passed as argument. But before calling this function it is checked whether feature id is none or not.

CHANGELOG.md
matthiasschaub: Added it to Breaking Changes section.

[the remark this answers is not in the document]
matthiasschaub: I removed the issue number in favor of the PR number.

workers/ohsome_quality_analyst/api.py
matthiasschaub: No, thats not quite right. The API parameter `bpolys` can be of type `Feature` or `Geometry`, but after parsing it using `load_bpolys()` it is always a `Feature` object.

matthiasschaub: I renamed the variable `bpolys` to `feature` after `load_bpolys()` has been called.

workers/ohsome_quality_analyst/base/indicator.py
matthiasschaub: See https://github.com/GIScience/ohsome-quality-analyst/pull/69#discussion_r664022120

workers/ohsome_quality_analyst/geodatabase/client.py
matthiasschaub: At least we expect only Polygon and MultiPolygon since OQT works on areas for now. Input features to OQT should have those types as geometry.
Also since we calculate the area only those two geometries are making sense.

workers/ohsome_quality_analyst/api.py
matthiasschaub: I changed all occurrences in api.py and oqt.py to `Optional[Union[int, str]]`

workers/ohsome_quality_analyst/api.py
```diff
@@ -104,14 +123,19 @@ async def _fetch_indicator(
     url: str,
     bpolys: Optional[str] = None,
     dataset: Optional[str] = None,
-    feature_id: Optional[Union[int, str]] = None,
+    feature_id: Union[int, str, None] = None,
```
joker234, line 126: https://github.com/GIScience/ohsome-quality-analyst/pull/69#discussion_r664030564

workers/ohsome_quality_analyst/api.py
joker234: I meant in general, why do you want a variable which is either `int` or `str`? In the database you can only handle one of them, right? So you can handle all identifier in OQT as string, or did I miss a problem with that? I think it would be much easier to focus on **one** type for identifier :wink:

workers/ohsome_quality_analyst/api.py
joker234: Okay I think you misread me here. In the past we allowed people to give OQT a GeoJSON with a FeatureCollection with **one** `Feature`. I keeping this could be fairly easy for API and CLI. It could be something like this:
```python
elif isinstance(bpolys, FeatureCollection):
    if len(bpolys.features) == 1:
        # either this one to use it recursive, or duplicate code by copying the Feature implementation
        return load_bpolys(geojson.drops(bpolys.features[0]))
    else:
        raise ValueError("Only one Feature is supported in a FeatureCollection")
```

workers/ohsome_quality_analyst/geodatabase/client.py
joker234: Is this documented somewhere? The whole “what can I pass to OQT” in which format should be documented somewhere. (GeoJSON with a Polygon or MultiPolygon, optionally within a Feature, optionally within a FeatureCollection).

workers/ohsome_quality_analyst/api.py
joker234: General question? Why is this implemented in `api.py` and not in `oqt.py`? Shouldn't the code be similar for both the API and CLI?

workers/ohsome_quality_analyst/geodatabase/client.py
```diff
@@ -143,51 +143,29 @@ async def get_area_of_bpolys(bpolys: Dict):
                     )
             ) / (1000*1000) as area_sqkm
         """
-    polygon = json.dumps(bpolys["features"][0]["geometry"])
     async with get_connection() as conn:
-        result = await conn.fetchrow(query, polygon)
+        result = await conn.fetchrow(query, geojson.dumps(bpolys))
```
joker234, line 147:
```suggestion
        # TODO when fixing the TODO to allow geojson.Feature: make sure the geometries (e.g. Polygon) are returned
        result = await conn.fetchrow(query, geojson.dumps(bpolys))
```

workers/ohsome_quality_analyst/geodatabase/client.py
```diff
@@ -143,51 +143,29 @@ async def get_area_of_bpolys(bpolys: Dict):
...
-                )
-            )
-        )
-        FROM {dataset}
-        WHERE {fid_field} = $1
-    """
-    ).format(fid_field=fid_field, dataset=dataset)
+        "SELECT ST_AsGeoJSON(geom) "
+        + "FROM regions "
+        + "WHERE {0} = $1".format(fid_field)
```
joker234, line 160: I think the `fid_field` can be directly passed from an API request to this point where the query is executed. We should have a sanity checking in the API implementation.

workers/ohsome_quality_analyst/indicators/ghs_pop_comparison_buildings/indicator.py
```diff
@@ -200,7 +199,6 @@ async def get_zonal_stats_population(self, bpolys: dict) -> Record:
                 st_setsrid(public.ST_GeomFromGeoJSON($3), 4326)
              )
             """
-        polygon = json.dumps(bpolys["features"][0]["geometry"])  # Geometry only
-        data = (polygon, polygon, polygon)
+        data = tuple(map(str, [self.feature.geometry] * 3))
```
joker234, line 202: The pythonic way would be something like this:
```suggestion
        data = (str(self.feature.geometry),) * 3
```
At least using a list and a map seems a little bit unnecessary. If you want to stick with the list, you can do something like this:
```suggestion
        data = tuple([str(self.feature.geometry)] * 3)
```

workers/ohsome_quality_analyst/indicators/ghs_pop_comparison_roads/indicator.py
```diff
@@ -197,7 +193,6 @@ async def get_zonal_stats_population(self, bpolys: dict) -> Record:
                 st_setsrid(public.ST_GeomFromGeoJSON($3), 4326)
              )
             """
-        polygon = json.dumps(bpolys["features"][0]["geometry"])  # Geometry only
-        data = (polygon, polygon, polygon)
+        data = tuple(map(str, [self.feature.geometry] * 3))
```
joker234, line 196: See comment in `ghs_pop_comparison_buildings/indicator.py`.

workers/tests/integrationtests/test_geodatabase.py
```diff
@@ -73,29 +73,20 @@ def test_get_feature_ids(self):
...
-                db_client.get_bpolys_from_db(self.dataset, self.feature_id, "foo")
-            )
+            asyncio.run(db_client.get_region_from_db(self.feature_id, "foo"))
 
         with self.assertRaises(DataError):
-            asyncio.run(
-                db_client.get_bpolys_from_db(self.dataset, "foo", self.fid_field)
-            )
-
-        with self.assertRaises(UndefinedTableError):
```
joker234, line 95: Did you remove this error? Or why don't we test this *here* anymore?

joker234 (commented):
Thanks for the PR and the additional tests :+1: I made some additional comments, but in general, the PR looks good.

workers/ohsome_quality_analyst/api.py
matthiasschaub: Ah, now I see what you mean. 
Since https://github.com/GIScience/ohsome-quality-analyst/pull/37 `feature_id` can be str or int depending on what `feature_id_field` is.

E.g. 
```bash
# int
oqt create-indicator --fid-field ogc_fid --feature-id 1
# str
oqt create-indicator --fid-field name --feature-id heidelberg
```

What do you think about this behavior?

workers/ohsome_quality_analyst/api.py
matthiasschaub: I added support for FeatureCollection with one Feature as input to API.

matthiasschaub: The goal is to share as much logic between API and CLI as possible as well as similar input/output. I hope to address this in future PRs. Right now the CLI supports GeoJSON FeatureCollection with more than one Feature. But as I said I hope to unify both of them in the near future.

workers/ohsome_quality_analyst/geodatabase/client.py
matthiasschaub: Currently it is not documented. It would be nice to document it before the release.

matthiasschaub: https://github.com/GIScience/ohsome-quality-analyst/issues/78

workers/ohsome_quality_analyst/api.py
matthiasschaub: Fixed

workers/ohsome_quality_analyst/cli.py
```diff
@@ -121,12 +120,11 @@ def create_indicator(
         # When using an infile as input
         feature_collection = load_infile(infile)
         for feature in feature_collection.features:
-            sub_collection = geojson.FeatureCollection([feature])
             indicator = asyncio.run(
                 oqt.create_indicator(
                     indicator_name,
                     layer_name,
-                    bpolys=sub_collection,
+                    feature=feature,
```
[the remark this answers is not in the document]
matthiasschaub: FeatureCollection with one Feature is now supported. See load_bpolys(), which returns the Feature from FeatureCollection of lenght one.

workers/ohsome_quality_analyst/cli.py
```diff
@@ -190,11 +184,10 @@ def create_report(
         # When using an infile as input
         feature_collection = load_infile(infile)
         for feature in feature_collection.features:
-            sub_collection = geojson.FeatureCollection([feature])
             report = asyncio.run(
                 oqt.create_report(
                     report_name,
-                    bpolys=sub_collection,
+                    feature=feature,
```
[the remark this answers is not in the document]
matthiasschaub: https://github.com/GIScience/ohsome-quality-analyst/pull/69#discussion_r664296896

workers/ohsome_quality_analyst/cli.py
```diff
@@ -223,12 +215,9 @@ def create_report(
...
                 fid_field = DATASETS[dataset_name]["default"]
-            feature_collection = asyncio.run(
-                db_client.get_bpolys_from_db(dataset_name, feature_id, fid_field)
-            )
-            for feature in feature_collection.features:
-                feature = update_features_report(feature, report)
-            write_geojson(outfile, feature_collection)
+            feature = asyncio.run(db_client.get_region_from_db(feature_id, fid_field))
+            feature = update_features_report(feature, report)
+            write_geojson(outfile, feature)
```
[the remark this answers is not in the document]
matthiasschaub: Before this PR `get_region_from_db()` returned a FeatureCollection with one Feature. I changed this to only return the Feature. 

If the parameter to the CLI are `dataset` and `feature_id` then OQT always act upon one feature only.
If on the other hand a GeoJSON is passed in then it can also be a FeatureCollection with multiple Features. See lines above at the beginning of the function (183).

workers/ohsome_quality_analyst/geodatabase/client.py
matthiasschaub: Not sure what you mean here. Do you expect that the function takes a Feature instead of Geomety as input?
What do you mean be returning the geometry in this case?

workers/ohsome_quality_analyst/geodatabase/client.py
matthiasschaub: fid_field is sanity checked in `oqt.py` using the function definied in `geodatabase/client.py`
```python
         if not db_client.sanity_check_fid_field(dataset, fid_field):
            raise ValueError("Input feature id field is not valid: " + fid_field)
```

workers/ohsome_quality_analyst/indicators/ghs_pop_comparison_buildings/indicator.py
matthiasschaub: I implemented your last suggestion

workers/ohsome_quality_analyst/indicators/ghs_pop_comparison_roads/indicator.py
matthiasschaub: I implemented your last suggestion

workers/tests/integrationtests/test_geodatabase.py
matthiasschaub: Now the function `get_bpolys_from_db()` operates only on the regions table. So this error can not be raised anymore. Before that when passing in an dataset which has not been present this error would have been raised.

matthiasschaub:
@joker234 Thanks for your attentive feedback! I addressed most of your feedback directly in the code. Some discussions are still open.

workers/ohsome_quality_analyst/geodatabase/client.py
joker234: That's what you (or someone else) put as a TODO above the function.
`# TODO Rewrite to work with geojson.Feature as input`

workers/ohsome_quality_analyst/geodatabase/client.py
joker234: missed that, thanks :)

workers/ohsome_quality_analyst/geodatabase/client.py
matthiasschaub: Ah yes, that was probably me. But I changed my mind on that. I think just the geometry is more appropriate.

joker234: then maybe delete the other TODO ;)

matthiasschaub: Sorry, I deleted the TODO comment

joker234 (approved):

workers/tests/integrationtests/test_geodatabase.py
```diff
@@ -100,6 +100,12 @@ def test_sanity_check_fid_field(self):
         self.assertFalse(db_client.sanity_check_fid_field(self.dataset, "foo"))
         self.assertTrue(db_client.sanity_check_fid_field(self.dataset, self.fid_field))
 
+    def test_type_of(self):
+        result = asyncio.run(db_client.type_of(self.dataset, self.fid_field))
+        self.assertEqual(result, "integer")
+        result = asyncio.run(db_client.type_of(self.dataset, "name"))
+        self.assertNotEqual(result, "integer")
```
joker234, line 107: do we need a (unit)test for that?

workers/ohsome_quality_analyst/geodatabase/client.py
```diff
@@ -190,3 +187,15 @@ def sanity_check_fid_field(dataset: str, fid_field: str) -> bool:
...
+async def type_of(table_name: str, column_name: str) -> str:
+    """Get data type of field"""
+    query = (
+        "SELECT data_type "
+        + "FROM information_schema.columns "
+        + "WHERE table_name = $1 AND column_name = $2"
+    )
+    async with get_connection() as conn:
+        record = await conn.fetchrow(query, table_name, column_name)
+    return record[0]
```
joker234, line 201: do we need a (unit)test for that?

joker234 (approved):

PR #101, opened 2021-07-21
Add osm-timestamps to indicators

Gigaszi:
### Description
Add osm-timestamps to indicators and add timezone to oqt-timestamp.

### Corresponding issue
Closes #1

### Checklist
- [x] I have updated my branch to `main` (e.g. through `git rebase main`)
- [x] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#pre-commit) before committing
- [x] I have commented my code
- [x] I have added sufficient unit and integration [tests](https://github.com/GIScience/ohsome-quality-analyst/blob/main/docs/development_setup.md#tests)
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CHANGELOG.md)

PR #113, opened 2021-07-28
Add api endpoints  list of indicators, reports, layers, datasets and fid_field

### LINKED ISSUE #106: Add API endpoints to retrive a list of indicators, reports, layers, datasets and fid_field
The API should have the same functionality as the CLI. See cli.py.

Gigaszi:
### Description
Add api-endpoints list of indicators, reports, layers, datasets and fid_field and tests for them.

### Corresponding issue
Closes #106 

-

### Checklist

Please check all finished tasks. If some tasks do not apply to your PR, please cross their text out (by using `~...~`) and remove their checkboxes.


CHANGELOG.md
```diff
@@ -11,11 +11,15 @@
 - Set proper User Agent for requests to the ohsome API ([#62])
 - Set log level matplotlib fontmanager to INFO ([#90])
 - Make VCR mode configurable ([#95])
+- Add osm-timestamps to indicators + tests for it. Add timezone to oqt-timestamp([#101])
```
joker234, line 14:
```suggestion
- Add osm-timestamps to indicators + tests for it. Add timezone to oqt-timestamp ([#101])
```

workers/ohsome_quality_analyst/api.py
```diff
@@ -211,6 +217,42 @@ async def _fetch_report(
     return response
 
 
-@app.get("/regions")
+@app.get("/list_regions")
```
joker234, line 220: Imho that's a breaking change. Please label/sort that accordingly in the CHANGELOG.md.

workers/ohsome_quality_analyst/api.py
```diff
@@ -211,6 +217,42 @@ async def _fetch_report(
...
 async def get_available_regions():
+    """List names of available regions."""
     return await db_client.get_available_regions()
+
+
+@app.get("/list_indicators")
+async def list_indicators():
```
joker234, line 227: As this is only returning the names of the indicators (in contrast to `/list_regions`), wouldn't it be better to name it to something like `/list_indicator_names`? // cc @matthiasschaub 

The same question applies to the other new endpoints.

workers/ohsome_quality_analyst/api.py
```diff
@@ -211,6 +217,42 @@ async def _fetch_report(
...
+    """List names of available indicators."""
+    return list(load_metadata("indicators").keys())
```
joker234, line 229: The return type should be `JSON` (or `dict` in python) as it is defined in the openapi.json.

See:
```bash
curl -s http://localhost:8080/openapi.json | python3 -m json.tool | grep "\/list_indicators" -A16
        "/list_indicators": {
            "get": {
                "summary": "List Indicators",
                "description": "List names of available indicators.",
                "operationId": "list_indicators_list_indicators_get",
                "responses": {
                    "200": {
                        "description": "Successful Response",
                        "content": {
                            "application/json": {
                                "schema": {}
                            }
                        }
                    }
                }
            }
        },
```
I would suggest something like this (@matthiasschaub please check if this response fits into the other API endpoints):
```suggestion
    response = empty_api_response()
    response["indicator_names"] = list(load_metadata("indicators").keys())
    return response
```

The same applies to the other new endpoints.

workers/tests/integrationtests/test_api_regions.py
```diff
@@ -11,16 +11,51 @@
...
+        url = "/list_regions"
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
         self.assertTrue(geojson.loads(response.content).is_valid)
 
+    def test_get_list_indicators(self):
+        url = "/list_indicators"
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertIsInstance(geojson.loads(response.content), list)
```
joker234, line 28: The response shouldn't be of type list as it is defined right now.

The same applies to the following tests.

joker234 (commented):
Thanks for the implementation! I made some comments :)

joker234:
I think you have in here changes that are actually done in #101 as well. Please check that they are not appearing in both PRs. Thanks :)

workers/ohsome_quality_analyst/api.py
matthiasschaub: Sounds reasonable to me. In this case I would change the commands of the CLI as well. On the other hand this leads to quite long commands with are tedious to write.

I would prefer shorter names with good description. The docstring of the function will also be displayed in the swagger documentation. Maybe this will be enough to avoid confusion. E.g. on the CLI side this looks like following:

```python
@cli.command("list-indicators")
def list_indicators():
    """List available indicators and their metadata."""
```

```bash
oqt --help
[...]
Commands:
  list-indicators        List available indicators and their metadata.
```

But I am open for changing the command/enpoint name.

matthiasschaub: Regarding the format. It should be formatted in lower camel case like the other parameters of the API. E.g. `listIndicators`. I find that hard to read and would prefer `list-indicators` as  this [article](https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/) suggests but this would be a change for another PR and needs to be discussed. Right now consistency is more important.

workers/ohsome_quality_analyst/api.py
matthiasschaub: I think you are right. The response should be a JSON (Python dict). Using the `emtpy_api_response` function for a template is a good idea, but the `attribution` field of the template is not needed and also not correct. Maybe writting two template functions would make sense.

```python
def api_resonse_base_template(request_url: str) -> dict:
    return {"apiVersion": oqt_version}


def api_response_template(request_url: Optional[str] = None) -> dict:
    template = api_resonse_base_template()
    if request_url is not None:
        template["requestUrl"] = request_url
    template["attribution"] = {
        "text": "© OpenStreetMap contributors",
        "url": "https://ohsome.org/copyrights",
    }
    return template.copy()

@app.get("/list_indicators")
async def list_indicators():
    """List names of available indicators."""
    response = api_resonse_base_template()
    response["indicator_names"] = list(load_metadata("indicators").keys())
    return response
```

matthiasschaub: At first I thought we could simply return a list, which is also a valid JSON response (JSON Array). But there are some reasons why this considered by some as no good practice. https://softwareengineering.stackexchange.com/questions/286293/whats-the-best-way-to-return-an-array-as-a-response-in-a-restful-api

PR #123, opened 2021-08-06
Fixing saturation indicator bug

### LINKED ISSUE #45: Different None types calculating the mapping saturation indicator
Calculating the Mapping Saturation Indicator, multiple expression for no data available/calculation not possible are used. The growth and saturation rate are sometimes labelled as "NULL" and sometimes as "nan". 

When "NULL" appears, the result label is always "undefined", however the result value is either "-1" or "Null".

When "nan" appears, the result label is always "yellow" and the result value is "0,5". In this case, the result description does not state that the indicator could not be calculated, but creates a wrong description in the way it does for succesful calculations (--> "with a growth rate of nan.")

NullNanBug

hn437:
Prevent the mapping saturation indicator from returning nan-Values as result if data is available. See Issue #45 for bug description

### Corresponding issue
Closes #45 

-

### Checklist

### DIFF
--- a/workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
+++ b/workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
@@ -1022,25 +1022,26 @@ class sigmoidCurve:
             "logistic4",
             "logistic5",
         ]
-
-        def check_no_nan_element(element):
-            return not math.isnan(element)
-
         # initial values for the single sigmoid curve
         initParamsSingle = self.initparamsingle(df1.li, df1.yValues)
         errorsListSingle = []
         if not math.isnan(initParamsSingle[3]):
             inits5curves = self.sortInits5curves(df1.li, df1.yValues)
+            # add up the 2 lists and then check each of the 4 values,
+            # if they are not nan
             validity_result = all(
-                map(
-                    check_no_nan_element,
-                    [
-                        math.fsum(inits5curves[0]),
-                        math.fsum(inits5curves[1]),
-                        inits5curves[2],
-                        inits5curves[3],
-                    ],
-                )
+                [
+                    not x
+                    for x in map(
+                        math.isnan,
+                        [
+                            math.fsum(inits5curves[0]),
+                            math.fsum(inits5curves[1]),
+                            inits5curves[2],
+                            inits5curves[3],
+                        ],
+                    )
+                ]
             )
             if validity_result is True:
                 # get possible xmids
@@ -1078,15 +1079,18 @@ class sigmoidCurve:
             errorsListSingle = []
             inits5curves = self.sortInits5curves(df1.li, df1.yValues)
             validity_result = all(
-                map(
-                    check_no_nan_element,
-                    [
-                        math.fsum(inits5curves[0]),
-                        math.fsum(inits5curves[1]),
-                        inits5curves[2],
-                        inits5curves[3],
-                    ],
-                )
+                [
+                    not x
+                    for x in map(
+                        math.isnan,
+                        [
+                            math.fsum(inits5curves[0]),
+                            math.fsum(inits5curves[1]),
+                            inits5curves[2],
+                            inits5curves[3],
+                        ],
+                    )
+                ]
             )
             if validity_result is True:
                 # get possible xmids
--- a/workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
+++ b/workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
@@ -1022,12 +1022,27 @@ class sigmoidCurve:
             "logistic4",
             "logistic5",
         ]
+
+        def check_no_nan_element(element):
+            return not math.isnan(element)
+
         # initial values for the single sigmoid curve
         initParamsSingle = self.initparamsingle(df1.li, df1.yValues)
         errorsListSingle = []
         if not math.isnan(initParamsSingle[3]):
             inits5curves = self.sortInits5curves(df1.li, df1.yValues)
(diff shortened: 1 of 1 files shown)


hn437:
Change to prevent the bug at the moment: Instead of checking a single parameter for nan-values, check all values within the 4 parameters for nan-values and use else-clause if nan-values do exist

joker234:
Is `inf` a valid outcome we want to allow? If not you could reduce your check to something like this:
```
all(map(math.isfinite, a))
```
with `a` as your input list

hn437:
> Is `inf` a valid outcome we want to allow? If not you could reduce your check to something like this:
> 
> ```
> all(map(math.isfinite, a))
> ```
> 
> with `a` as your input list

yes, but you need to check 2 lists and 2 floats which are all written in one tuple. therefore you need to handle the first to elements differently than the last two. and you need to check the results for each element within the nested list as well as the other float elements in the tuple

hn437:
> > Is `inf` a valid outcome we want to allow? If not you could reduce your check to something like this:
> > ```
> > all(map(math.isfinite, a))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with `a` as your input list
> 
> no, but you need to check 2 lists and 2 floats which are all written in one tuple. therefore you need to handle the first to elements differently than the last two. and you need to check the results for each element within the nested list as well as the other float elements in the tuple

so you could replace it with this
`
validity_result = all(
                [
                    all(map(math.isfinite, inits5curves[0])),
                    all(map(math.isfinite, inits5curves[1])),
                    math.isfinite(inits5curves[2]),
                    math.isfinite(inits5curves[3]),
                ]
            )
`

is this prefered?

joker234:
> …
> so you could replace this:
> `validity_result = all( [ not x for x in map( math.isnan, [ math.fsum( inits5curves[0] ), math.fsum(inits5curves[1]), inits5curves[2], inits5curves[3], ], ) ] )`
> with this
> `validity_result = all( [ all(map(math.isfinite, inits5curves[0])), all(map(math.isfinite, inits5curves[1])), math.isfinite(inits5curves[2]), math.isfinite(inits5curves[3]), ] )`
> 
> is this prefered?

I meant something like this, it's just a little bit shorter and you don't have to use that much nesting layers ;)
```python
validity_result = all(map(math.isfinite, [*inits5curves[0] *inits5curves[1], inits5curves[2], inits5curves[3]]))
```

hn437:
> > …
> > so you could replace this:
> > `validity_result = all( [ not x for x in map( math.isnan, [ math.fsum( inits5curves[0] ), math.fsum(inits5curves[1]), inits5curves[2], inits5curves[3], ], ) ] )`
> > with this
> > `validity_result = all( [ all(map(math.isfinite, inits5curves[0])), all(map(math.isfinite, inits5curves[1])), math.isfinite(inits5curves[2]), math.isfinite(inits5curves[3]), ] )`
> > is this prefered?
> 
> I meant something like this, it's just a little bit shorter and you don't have to use that much nesting layers ;)
> 
> ```python
> validity_result = all(map(math.isfinite, [*inits5curves[0] *inits5curves[1], inits5curves[2], inits5curves[3]]))
> ```

Okay, thank you! Didn't knew you could use the * like that... implemented it!

workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
```diff
@@ -1027,7 +1027,20 @@ def getBestFittingCurve(self, preprocessing_results: Dict) -> list:
         errorsListSingle = []
         if not math.isnan(initParamsSingle[3]):
             inits5curves = self.sortInits5curves(df1.li, df1.yValues)
-            if not math.isnan(inits5curves[3]):
+            # add up the 2 lists and then check each of the 4 values,
+            # if they are not nan
+            validity_result = all(  # check if all elements are true
+                map(
+                    math.isfinite,  # check if number is finite --> excludes nan
```
matthiasschaub, line 1034: Why do you use `isfinite` instead of the more explicit and easier to understand `ìsnan`?

workers/tests/integrationtests/test_indicator_mapping_saturation.py
```diff
@@ -57,6 +58,26 @@ def test_cannot_convert_nan_error(self):
         indicator.calculate()
         indicator.create_figure()
 
+    @oqt_vcr.use_cassette()
+    def test_avoiding_nan_in_sigmoid_curve(self):
+        """In some areas which contain mapped features, but mapping has happened in very
+        few contributions, sometimes not all curves could be calculated correctly and
+        those were chosen as the best-fitting curve. As a result, the output contained
+        nan-values instead of data or None values. This test checks the fix which avoids
+        choosing a curve containing nan values"""
```
matthiasschaub, line 67: I would try to write pep conform docstrings: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. 

```suggestion
        """Test for NaN values of the `saturation` attribute.
        
        In some areas which contain mapped features, but mapping has happened in very
        few contributions, sometimes not all curves could be calculated correctly and
        those were chosen as the best-fitting curve. As a result, the output contained
        nan-values instead of data or None values. This test checks the fix which avoids
        choosing a curve containing nan values.
        """
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py
matthiasschaub: Sorry, missed previous discussion about this change.

matthiasschaub (approved):

PR #170, opened 2021-09-10
Rewrite of the Mapping Saturation indicator

### LINKED ISSUE #175: Mapping Saturation Indicator results should include best fitting curve description
Once a best fitting curve is found (sigmoid with two jumps or sigmoid with three jumps etc.) the returned result of the indicator should contain which one it is.

- [x] Add a new attribute the Mapping Saturation e.g: `best_fitting_curve: str = "double sigmoid"`

matthiasschaub:
### Description

- Use scipy to fit sigmoid model to data
- Implement additional statistical models using R
- Implement additional statistical metrics (errors)

Implement additional regression models (Functions for fitting optimal curves) for the Mapping Saturation indicator. The usage of those models are based on current research. R and its stats module is used to run those models. Communication between Python and R is done by the Python package `rpy2`. The following models are all Self-Starting. That means the initial guess is taking care of by the model itself.

The 5 additional models implemented are:
- Self-Starting Nls Michaelis-Menten Model: `Vm * x / (K + x)`
- Self-Starting Nls Asymptotic Regression Model: `asym + (R0 - asym) * e^(-e^(lrc) * x)`
- Self-Starting Nls Four-Parameter Logistic Model: `A + (B - A) / (1 + e^((xmid - x) / scal))`
- Self-Starting Nls Logistic Model: `asym / (1 + e^((xmid - x) / scal))`
- Two-Steps-Sigmoidal / Tangens Hyperbolicus: (See below)
```
e + (f - e) * 1 / 2 * (np.tanh(k * (x - b)) + 1) + (Z - f) * 1 / 2 * (np.tanh(k * (x - c)) + 1)
```

uml-class-diagram

### Corresponding issue

Closes #175 #198 #208 #217

- Fix Python version to 3.8>3.11 (`pyproject.toml` and `Dockerfile`)
- System Package: `R`
- Python: 
  - `rpy2`
  - `scipy`

### Checklist

- [x] I have rebased branch to make commit history clearer
- [x] Check if all relevant attributes of issue #175 are included in the API response 
- [x] Check if result.value reflects issues raised in #198 
- [x] Merge with PR #209 
    - [x] Update PR description to reflect changes implemted in PR #209 
- [x] Reference paper
    - Update indicator description in `indicator.py` and `metadata.yaml`
    - Additional reference paper will be added once in is published.
- [x] Add documentation about the integration of R


matthiasschaub:
Output of `sigmoid_curve_2` for comparison with original results (`sigmoid_curve`):

Rio De Janeiro
Athen
Kangal
Saterland

matthiasschaub:
new-fid14
old-fid14
new-fid03
old-fid03
old-fid09
new-fid09

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -0,0 +1,103 @@
...
+    k = []
+    l = []  # noqa
+    for i in range(n):
+        x_0.append((len(xdata) / (n + 1) * (i + 1)))
+        k.append(0)
+        l.append((max(ydata) / n) * (i + 1))
+    return tuple(x_0 + k + l)
+
+
+def get_bounds(n: Literal[1, 2, 3, 4], xdata: list, ydata: list) -> Tuple[tuple]:
```
joker234, line 75: Your typing seems to be wrong. Correct would be one of these:
* `tuple`
* `Tuple[tuple, tuple]`
* `Tuple[Tuple[int], Tuple[float]]` ← I'm not sure if `int` shouldn't be float here. This needs to be fixed in the result, probably.

joker234 (commented):
Gone over most of your changes. Thanks for them :+1: 
I have to have a deeper look into the changes in `indicator.py`, but in general they look good as well.

workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curves.py
```diff
@@ -0,0 +1,59 @@
...
+    Args:
+        l: the curve's maximum value / asymptotic (the plateaus);
+        k: the logistic growth rate or steepness of the curve
+        x_0: the x value of the sigmoid's midpoint (inflection point)
+    """
+    return l / (1 + np.exp(-k * (x - x_0)))
+
+
+# Flake8 E741: Ambiguous variable name 'l'
+def sigmoid_1(x, x_0, k, l):  # noqa E741
```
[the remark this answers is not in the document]
matthiasschaub: I find the current implementation more clearer. E.g. It is easier to figure out what the math formula of a double sigmoid curve is. And quite easy to see that we will only combine up to four sigmoid curves.

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
matthiasschaub: Thanks for pointing it out. I corrected it to:
`Tuple[Tuple[int], Tuple[float]]`

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -0,0 +1,103 @@
...
+            continue
+        ydata_fitted = func(xdata, *popt)
+        mse = calc_mse(ydata, ydata_fitted)
+        fit = Fit(func_name, ydata_fitted, mse)
+        if best_fit is None or best_fit.mse > fit.mse:
+            best_fit = fit
+    return best_fit
+
+
+def get_initial_guess(n: Literal[1, 2, 3, 4], xdata: list, ydata: list) -> tuple:
```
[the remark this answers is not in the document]
matthiasschaub: Oh, thats good to now! I changed it to be of type int, added docs and assert statement.

workers/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.8.9
```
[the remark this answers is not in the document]
matthiasschaub: I will check again if it works without.

workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curves.py
```diff
@@ -0,0 +1,59 @@
+"""Sigmoid functions with up to four jumps."""
+
+import numpy as np
+
+
+# Flake8 E741: Ambiguous variable name 'l'
+def sigmoid(x: float, x_0: float, k: float, l: float):  # noqa E741
```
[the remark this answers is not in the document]
matthiasschaub: Yes I have seen this suggestion in PEP8.
But when `L` is used as single capital variable name two more violations will arise: 
- N803: function name should be lowercase (function names)
- N806: variable in function should be lowercase

That was how I first implemented it. I would like to capitialize the variable name. Then above erros should be added to Flake8 and SonarCloud configuration to be ignored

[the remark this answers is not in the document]
joker234: True, both are not ideal. But since the original variable name is a capitalized `L` and you have to ignore some rules for both cases. I would say the “don't use a lowercase `l` because you can mix it up with other letters” argument is the stronger one here.

workers/Dockerfile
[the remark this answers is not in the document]
joker234: We should never depend on a patch-level version. If really necessary, you could depend on one with `^3.8.9` (not here, but in the `pyproject.toml`). And you shouldn't use a version that is not available in normal* Linux distributions.

*normal as in non-cutting-edge versions

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -0,0 +1,105 @@
...
+        Each element of the tuple is an list with the length equal to the number
+        of parameters.
+    """
+    assert n in (1, 2, 3, 4)
+    x_0_upper_bounds = [len(xdata) * 1.5] * n
+    x_0_lower_bounds = [0.0] * n
+    k_upper_bounds = [1.0] * n
+    k_lower_bounds = [-1.0] * n
+    L_upper_bounds = [(max(ydata) / n) * (i + 1.0) for i in range(n)]  # noqa N806
+    L_lower_bounds = [0.0] * n  # noqa N806
```
joker234, line 96: if you want to use the capitalized `L` here as well, I would suggest to ignore the line for sonar:
```suggestion
    L_upper_bounds = [(max(ydata) / n) * (i + 1.0) for i in range(n)]  # noqa N806 NOSONAR
    L_lower_bounds = [0.0] * n  # noqa N806 NOSONAR
```

workers/pyproject.toml
```diff
@@ -15,7 +15,7 @@ keywords = [
   ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "3.8.9"
```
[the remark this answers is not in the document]
matthiasschaub: I changed the version number to `~3.8`, which alows only patch-level.

workers/ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curves.py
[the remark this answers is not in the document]
matthiasschaub: Capital `L` is now used instead of `l`

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
matthiasschaub: Added NOSONAR comments

docs/development_setup.md
```diff
@@ -47,8 +47,9 @@ A minimal database setup for running tests is provided. If the build argument `O
 
 ### Requirements
 
-- Python 3.8+
-- Poetry 1.1.0+
+- Python: ~3.8
+- R: ~4.0
+- Poetry: ~1.0
```
joker234, line 52: An approximate version isn't helpful. We should have a minimum version or version range.

workers/Dockerfile
```diff
@@ -1,5 +1,10 @@
+# based on Debian 11 Bullseye
 FROM python:3.8
 
+# install R
+RUN apt-get update
```
joker234, line 5: Please don't include the apt cache in image layers.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Improve API response when an error occurs by including information about the cause ([#186])
 - Add parameter to get API response without svg string in the result ([#137])
 - Extend time range for Currentness indicator ([#178])
+- Rewrite the Mapping Saturation indicator to use statistical models from the R language ([#209] [#170])
```
joker234, line 19: - [ ] this change isn't in 0.7.0
- [ ] please add a line to “How to upgrade”
- [ ] `[#209]` missing
- [ ] consistent would be:
```suggestion
- Rewrite the Mapping Saturation indicator to use statistical models from the R language ([#170], [#209])
```

CHANGELOG.md
```diff
@@ -27,11 +28,13 @@
     - E.g. `/indicator?name=Currentness`
 
 [#137]: https://github.com/GIScience/ohsome-quality-analyst/issues/137
+[#170]: https://github.com/GIScience/ohsome-quality-analyst/pull/170
```
joker234, line 31: no link in `0.7.0`

CHANGELOG.md
```diff
@@ -27,11 +28,13 @@
...
 [#178]: https://github.com/GIScience/ohsome-quality-analyst/pull/178
 [#186]: https://github.com/GIScience/ohsome-quality-analyst/pull/186
 [#190]: https://github.com/GIScience/ohsome-quality-analyst/pull/190
 [#199]: https://github.com/GIScience/ohsome-quality-analyst/pull/199
 [#204]: https://github.com/GIScience/ohsome-quality-analyst/pull/204
+[`scipy`]: https://scipy.org/
```
joker234, line 37: Why is this here?

workers/setup.cfg
```diff
@@ -12,13 +11,14 @@ inline-quotes = double
 max-line-length = 88
 
 per-file-ignores =
-    # TODO: Fix issues causing those warnings
-    ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py:N806,N802,N803,N801
+    # N806: variable in function should be lowercase
```
joker234, line 14: Why is this line at the top?

workers/pyproject.toml
```diff
@@ -15,7 +15,7 @@ keywords = [
   ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "~3.8"
```
joker234, line 18: Why this change? All the other dependencies are defined with the caret and not tilde. Is there an issue with Python 3.9?

workers/pyproject.toml
```diff
@@ -30,6 +30,8 @@ httpx = "^0.17.0"
 asyncpg = "^0.22.0"
 vcrpy = "^4.1.1"
 python-dateutil = "^2.8.2"
+scipy = "^1.7.1"
```
joker234, line 33: Is there a reason you don't use the most recent version?

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,157 +35,115 @@ def __init__(
...
+            except RRuntimeError as error:
+                logging.warning(
+                    "Could not run model {0} due to RRuntimeError: {1}".format(
+                        model.name, error
+                    )
+                )
+                continue
+        self.best_fit = min(self.fitted_models, key=lambda m: m.mae)
+        logging.info("Best fitting model: " + self.best_fit.name)
+        self.saturation = (
```
Hagellach37, line 94: I think from here on we need to double check how we would like to approach this now. In the past we based our quality label on the saturation level. However, we could now change this and rather rely on the asymptote value of the fitted function.

If we decide to go for the asymptote approach we need to change the thresholds. The good thing would be that we can derive self.value as the completeness

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -194,3 +152,29 @@ def create_figure(self) -> None:
         self.result.svg = img_data.getvalue()  # this is svg data
         logging.debug("Successful SVG figure creation")
         plt.close("all")
+
+    def check_corner_cases(self) -> bool:
```
Hagellach37, line 156: maybe not so important, but I think `edge_case` might be the better word? 

* https://en.wikipedia.org/wiki/Edge_case
* https://en.wikipedia.org/wiki/Corner_case

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -194,3 +152,29 @@ def create_figure(self) -> None:
...
+        """Check corner cases
+
+        If corner case is present set result description.
+
+        Returns
+            bool: Return True if no corner case present. False otherwise.
```
Hagellach37, line 162: I find this a bit counter intuitive. You check for a certain condition (e.g. if this is a edge case) and return True if it is not present. I would have expected that you return True if it is a corner case.

Or did you have a reason to do it the other way round?

workers/ohsome_quality_analyst/indicators/mapping_saturation/ssdoubles.R
```diff
@@ -0,0 +1,41 @@
+#############################
+## Self-starting two-steps-sigmoidal function
+## TODO: Reference Authors
```
Hagellach37, line 3: this is the function implement by Josephine? I think it's good to have it here in R.

workers/pyproject.toml
matthiasschaub: With `~3.8` I got following error when running `poetry`.
```
  The current project's Python requirement (>=3.8,<4.0) is not compatible with some of the required packages Python requirement:
    - scipy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0

  Because scipy (1.7.3) requires Python >=3.7,<3.11
   and no versions of scipy match >1.7.3,<2.0.0, scipy is forbidden.
  So, because ohsome-quality-analyst depends on scipy (^1.7.3), version solving failed.
  
  [ ... ]
  
  
  ```
  So I though it would be a good idea to fix the version to a specific minor version. But now I chose to use following version requirement:

Comment (matthiasschaub) [reply workers/pyproject.toml:18]: With `~3.8` I got following error when running `poetry`.
```
  The current project's Python requirement (>=3.8,<4.0) is not compatible with some of the required packages Python requirement:
    - scipy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0

  Because scipy (1.7.3) requires Python >=3.7,<3.11
   and no versions of scipy match >1.7.3,<2.0.0, scipy is forbidden.
  So, because ohsome-quality-analyst depends on scipy (^1.7.3), version solving failed.
  
  [ ... ]
  
  ```
  So I though it would be a good idea to fix the version to a specific minor version. But now I chose to use following version requirement:

workers/pyproject.toml
matthiasschaub: Not sure why this came about. I will update to the most recent version.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
matthiasschaub: I changed the word to edge case.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
matthiasschaub: Yes I understand this. I changed the behavior of the function to return a string with the description of the edge case if present. I think now it is easier to understand.

workers/Dockerfile
matthiasschaub: Thanks for pointing out. I will make sure not to include the apt-cache in the image layer by removing it after installation.

matthiasschaub: Thanks for pointing out. I will make sure not to include the apt-cache in the image layer by removing it after installation.

matthiasschaub:
For all OQT regions a total number of 192 models are expected. But only 112 models could be created.
A RRuntimeError with an error message will be raised if a model could not be created. Below is a collection of those error message for each model.

- Self-Starting Nls Asymptotic Regression Model
  - step factor 0.000488281 reduced below 'minFactor' of 0.000976562
  - singular gradient
- Self-Starting Nls Michaelis-Menten Model
  -  NA/NaN/Inf in 'y'
  - step factor 0.000488281 reduced below 'minFactor' of 0.000976562
- Two-Steps-Sigmoidal Model (Tangens Hyperbolicus)
  - singular gradient matrix at initial parameter estimates
- Self-Starting Nls Four-Parameter Logistic Model
  - step factor 0.000488281 reduced below 'minFactor' of 0.000976562
  - number of iterations exceeded maximum of 50
  - singular gradient
  - NA/NaN/Inf in foreign function call (arg 1)
- Self-Starting Nls Logistic Model due to RRuntimeError
  - step factor 0.000488281 reduced below 'minFactor' of 0.000976562
  - NA/NaN/Inf in foreign function call (arg 1)

docs/development_setup.md
matthiasschaub: I used those notation because they are used by poetry. But I recognize that this is confusing. I changed it to use `<` and `>`.

CHANGELOG.md
matthiasschaub: Its from an old sentence in the changelog. I removed this line.

workers/setup.cfg
matthiasschaub: You mean the explanation of the `N806` error? Its there so one knows what kind of issues are ignored per file by flake8.

workers/ohsome_quality_analyst/indicators/mapping_saturation/ssdoubles.R
matthiasschaub: Yes, this is from Josephine. I added her name as author to this R-file.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,408 @@
...
+                "mae",
+                "xdata",
+                "ydata",
+                "coefficients",
+                "fitted_values",
+            )
+        }
+
+
+class Sigmoid(BaseStatModel):
```
Hagellach37, line 64: I'm not sure if this model always returns what we expect. E.g. for Athens the plot looks like this. Are you sure that the asymptote is correct? From the plot I would expect that it could be higher, but maybe I'm wrong.

image

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,408 @@
...
+        alpha = 1.0 - level
+        degrees_of_freedom = len(self.ydata) - 3  # 3 is the number of parameters
+        tval = t_distribution.ppf(1.0 - alpha / 2.0, degrees_of_freedom)
+        perr = np.sqrt(np.diag(pcov))  # standard deviation errors
+        lower = popt - perr * tval
+        upper = popt + perr * tval
+        return np.array([lower[par_pos], upper[par_pos]])
+
+
+class SSlogis(BaseStatModel):
```
Hagellach37, line 162: also here you could double check if this result is realistic for Touggourt (Feature ID 4 in our regions on the website).

Again I would expect the asymptote to be higher.

image

Hagellach37 (commented):
Hey @matthiasschaub ,
nice work on the mapping saturation indicator. It's so much easier to understand now compared to our old code. There are few easy things to improve among my comments, e.g. legend placement in the plot, removing unused completeness attribute and fixing result description string values.

A bit more difficult might be the comments about how realistic the model results are. I think I might be good if we can have a discussion on that and check the results for more regions. I'm currently not 100% sure if the unexpected results are due to the mapping saturation approach itself or if there is some bug in the code.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +36,177 @@ def __init__(
...
+        self.upper_threshold = 0.95
+        self.lower_threshold = 0.5
 
         # Attributes needed for result determination
-        self.saturation = None
-        self.growth = None
+        self.best_fit: Optional[models.BaseStatModel] = None
+        self.fitted_models: List[models.BaseStatModel] = []
+        self.saturation: Optional[float] = None
+        self.completeness: Optional[float] = None
```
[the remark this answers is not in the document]
matthiasschaub: Ah, yes you are right. I removed the attribute.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +36,177 @@ def __init__(
...
-            # if saturation == 1.0:
-            #    growth should be 0.0
-            self.growth = 1 - self.saturation
-
-            description = Template(self.metadata.result_description).substitute(
-                saturation=self.saturation,
-                growth=self.growth,
+        if self.result.value >= self.upper_threshold:
+            self.result.label = "green"
+            self.result.description = (
```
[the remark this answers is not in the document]
matthiasschaub: Good catch. I fixed this by relying only on the result description and removing the description for each label. I think it is not necessary to provide a different description/interpretation for each label since it is a simple ratio in %. What do you think?

E.g. the description now is simply: Thea rea of interest is xx % complete.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +36,177 @@ def __init__(
...
-                "results": self.values,
-                "results_normalized": self.values_normalized,
-            }
+        self.best_fit = min(self.fitted_models, key=lambda m: m.mae)
+        logging.info("Best fitting model: " + self.best_fit.name)
+        # Calculate saturation: The slope of the last 3 years of the fitted curve
+        self.saturation = (
+            np.interp(xdata[-36], xdata, self.best_fit.fitted_values)
+        ) / (np.interp(xdata[-1], xdata, self.best_fit.fitted_values))
+        self.result.value = self.values[-1] / self.best_fit.asymptote  # Completeness
```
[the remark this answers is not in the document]
matthiasschaub: I added a new attribute called `latest_value`. I do not want to call it latest_feature_count or similar since it is not always a count (e.g. road length)

workers/Dockerfile
```diff
@@ -1,5 +1,13 @@
+# based on Debian 11 Bullseye
 FROM python:3.8
```
joker234, line 2: This comment isn't true for the future, isn't it? If you want to have Debian bullseye, you have to use `debian:bullseye` instead of `python:3.8`.

docs/development_setup.md
```diff
@@ -47,8 +47,9 @@ A minimal database setup for running tests is provided. If the build argument `O
 
 ### Requirements
 
-- Python 3.8+
-- Poetry 1.1.0+
+- Python: 3.8 > 3.11
```
joker234, line 50: Version 3.8 isn't larger than 3.11 ;) I think you wanted to write something like this: `Python: ≥ 3.8 and ≤ 3.11`
Why is this even important? Can't you just say ≥ 3.8?

docs/development_setup.md
```diff
@@ -47,8 +47,9 @@ A minimal database setup for running tests is provided. If the build argument `O
...
+- R: ^4.0
+- Poetry: ^1.0
```
joker234, line 52: I wouldn't use `^` in a manual. I would suggest to use human readable Syntax, e.g. using `≥` or `4.x`.

joker234, line 52: Why the downgrade in the Poetry version? Wasn't there a reason we required 1.**1**.0 or larger?

joker234 (commented):
Just some first comments/questions:
* Is there any documentation on the R integration? I couldn't find anything in the changes. Ideally this could be linked at an appropriate place if not even from the Changelog itself.
* Why is Jenkins running without changing the `Jenkinsfile`? Are the tests are running or already written?

docs/development_setup.md
matthiasschaub: Thanks for the feedback on the version declaration. I will integrate your comments.

Why not > 3.8? Currently in the `pyproject.toml` the Python version is fixed to `>=3.8,<3.11` and not `^3.8`. That is because of the package `scipy`. `poetry` will throw an install error if one tries to install `scipy` with the Python version fixed to `^3.8`. `poetry` suggest to use `>=3.8,<3.11` instead to resolve this issue.

docs/development_setup.md
matthiasschaub: Yes, thats easier to read. Will fix it.

matthiasschaub: Good Point. Not sure about the reason but I will revert the change.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: That is an interesting case. Here are all the models which produced fit.

Figure_2

My guess is that because of the step wise nature of the graph the curve fitting is difficult for the models. Since initial guess and bounds for the curve fitting is done by the R model itself there is not much which can be done about that. Except maybe to introduce weights to make the last data-points more important.

Another interesting edge case is that one would expect to rule this model out since the data is above the asymptote (more then the 5% margin). But the actual criterion here is  "if the latest data (average last two years) is lower than the upper limit of the 95% confidence interval of the estimated asymptote"

matthiasschaub: Somehow the graphs I posted are not showing the same data as the graph you posted. But I am pretty sure that my graphs are showing the result for the Touggourt region. Let me do some more checks.

matthiasschaub: @Hagellach37, your screenshot is based on an old commit where the asymptote has not been calculated correctly. Here is the current version (the percentage number format will be fixed). 
screenshort

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: For context here are the plots of all the models which produced fits:
Figure_11

matthiasschaub: From the above figure the plot with the smallest Mean Absolute Error is in the middle.
Using the latest commit this is also the plot which will be choosen and displayed on the website.
screenshort1

workers/Dockerfile
matthiasschaub: I removed the Debian version from the comment.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
Hagellach37: I think your plot does not refer to the same regions a my plot. My plot was for Athens.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
Hagellach37: I can't reproduce your figure. for me the completeness is 113%

image

Hagellach37: I'm a bit confused why your plot has a maximum value of around 200 for the OSM data and my plot shows a value higher than 350. Are we using the same region?

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: I just double checked using the website and the script for creating the plots and the above plots do refer to region of Athen (for me it is fid=1)

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +37,173 @@ def __init__(
...
-            {
-                "timestamps": self.timestamps,
-                "results": self.values,
-                "results_normalized": self.values_normalized,
-            }
+        self.best_fit = min(self.fitted_models, key=lambda m: m.mae)
+        logging.info("Best fitting model: " + self.best_fit.name)
+        # Slope of the last 3 years of the fitted curve
+        slope, *_ = linregress(xdata[-36:], self.best_fit.fitted_values[-36:])
+        self.result.value = round(slope, 2)
```
matthiasschaub, line 104: Slope is not a value between 0 and 1 and does not reflect saturation.

docs/development_setup.md
```diff
@@ -54,8 +54,9 @@ A minimal database setup for running tests is provided. If the build argument `O
 
 ### Requirements
 
-- Python 3.8+
-- Poetry 1.1.0+
+- Python: ≥ 3.8 and ≤ 3.11
+- R: ≥ 4.0
```
joker234, line 58: I don't this this is sufficient, please have a look at the [CI results](https://jenkins.ohsome.org/blue/organizations/jenkins/ohsome-quality-analyst/detail/PR-170/109/pipeline).

workers/Dockerfile
```diff
@@ -1,5 +1,13 @@
+# based on Debian
 FROM python:3.8
 
+# install R
+# to avoid caching issues combine apt-get update and install in one RUN statement.
+# to reduce image size, clean up the apt cache by removing /var/lib/apt/lists.
+RUN apt-get update && \
+    apt-get install -y r-base && \
+    rm -rf /var/lib/apt/lists/*
```
joker234, line 9: I don't this this is sufficient, please have a look at the [CI results](https://jenkins.ohsome.org/blue/organizations/jenkins/ohsome-quality-analyst/detail/PR-170/109/pipeline).

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+. All those models are self-starting nonlinear models. Nonlinear least-squares method is
+used to estimate the parameters.
+
+The package `rpy2` is an interface to R running embedded in a Python process:
+https://rpy2.github.io/
+
+Self-starting means the model will make the intial guess. Non-linear fits are sensitive
+to the initial guess. If the inital guess is not good following error can get raised by
+R: "singular gradient matrix at initial parameter estimates". In this case `rpy2` will
+raise an `RRuntimeError`.
```
joker234, line 13:
```suggestion
Self-starting means the model will make the initial guess. Non-linear fits are sensitive
to the initial guess. If the initial guess is not good following error can get raised by
R: "singular gradient matrix at initial parameter estimates". In this case `rpy2` will
raise an `RRuntimeError`.
```

docs/development_setup.md
matthiasschaub: I think it does. The CI failed because of some kind of `httpx` connection error during the run of `tests/integrationtests/test_indicator_ghs_pop_comparison_roads.py`. Not sure why there needs to be an `httpx` connection to the internet during this test but I will look into this.

Regarding the logging errors at the end of the pytest run (which are related to the rpy2 and pytest packages): There exists an issue to discuss this. #217

matthiasschaub: The avove mentioned integration test is missing the vrc decorator and hence will need to contact the ohsome API. I will make a PR to fix this.

workers/pyproject.toml
joker234: maybe add a comment to it, then :white_check_mark:

matthiasschaub: Added a comment to pyproject.toml.

workers/setup.cfg
```diff
@@ -12,13 +11,14 @@ inline-quotes = double
 max-line-length = 88
 
 per-file-ignores =
-    # TODO: Fix issues causing those warnings
-    ohsome_quality_analyst/indicators/mapping_saturation/sigmoid_curve.py:N806,N802,N803,N801
     # N803: argument name should be lowercase
-    #   Ignore for Fast-API parameters which are defined as mixedCase
     # N815: mixedCase variable in class scope (constants, method names)
```
joker234, line 15:
```suggestion
```

workers/setup.cfg
```diff
@@ -12,13 +11,14 @@ inline-quotes = double
...
-    #   Ignore for Fast-API parameters which are defined as mixedCase
+    # N806: variable in function should be lowercase
+    # N815: mixedCase variable in class scope (constants, method names)
+    # Ignore for Fast-API parameters which are defined as mixedCase
     ohsome_quality_analyst/api/api.py:N803,N815
```
joker234, line 19: Is N815 still necessary?
```suggestion
    # Ignore for Fast-API parameters which are defined as mixedCase
    ohsome_quality_analyst/api/api.py:N803
```

workers/ohsome_quality_analyst/api/api.py
```diff
@@ -161,9 +168,7 @@ async def _fetch_indicator(parameters) -> dict:
         remove_svg_from_properties(geojson_object)
     response = empty_api_response()
     response.update(geojson_object)
-    return JSONResponse(
-        content=jsonable_encoder(response), media_type=MEDIA_TYPE_GEOJSON
-    )
+    return CustomJSONResponse(content=response, media_type=MEDIA_TYPE_GEOJSON)
```
joker234, line 171: doesn't return a `dict` as indicated in the type hints

joker234 (commented):
- [ ] [api/api.py#195](https://github.com/GIScience/ohsome-quality-analyst/blob/7f760d5e0651ce6d5ae814c96eb59fd9227d7edb/workers/ohsome_quality_analyst/api/api.py#L195) expects a `dict`, not `ReportPOSTRequestModel`
- [ ] [api/api.py#196](https://github.com/GIScience/ohsome-quality-analyst/blob/7f760d5e0651ce6d5ae814c96eb59fd9227d7edb/workers/ohsome_quality_analyst/api/api.py#L196) `.dict()` doesn't make sense for type `dict`

workers/ohsome_quality_analyst/geodatabase/client.py
```diff
@@ -120,7 +118,7 @@ async def load_indicator_results(indicator, dataset: str, feature_id: str) -> bo
     indicator.result.svg = query_result["result_svg"]
 
     feature = geojson.loads(query_result["feature"])
-    properties = unflatten_dict(feature["properties"])
+    properties = unflatten_dict(feature["properties"], reverse=True)
```
joker234, line 121: As mentioned before, I think we don't need a parameter like that.

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +45,183 @@ def __init__(
...
         # Attributes needed for calculation
         self.values: list = []
-        self.values_normalized: list = []
+        self.latest_value: Union[None, int, float] = None
         self.timestamps: list = []
-        self.no_data: bool = False
-        self.deleted_data: bool = False
+
+        self.upper_threshold = 0.97  # Threshold derived from Gröchenig et al.
+        # TODO: What is a good lower threshold?
```
joker234, line 57: Is this already an issue somewhere?

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +45,183 @@ def __init__(
...
+            self.values.append(item["value"])
+            self.timestamps.append(isoparse(item["timestamp"]))
+        self.latest_value = self.values[-1]
 
     def calculate(self) -> None:
-        """Calculate the growth rate and saturation level within the last 3 years."""
-        if self.no_data:
-            self.result.description = "No features were mapped in this region."
+        # Latest timestamp of ohsome API results
+        self.result.timestamp_osm = self.timestamps[-1]
```
joker234, line 77: General question: where is documented what `result.timestamp_osm` is? Shouldn't this be in the documentation or the base indicator?

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +45,183 @@ def __init__(
...
-                # getSaturationInLast3Years returns 1.0
-
-            # if saturation == 1.0:
-            #    growth should be 0.0
-            self.growth = 1 - self.saturation
-
-            description = Template(self.metadata.result_description).substitute(
-                saturation=self.saturation,
-                growth=self.growth,
+        if self.result.value <= 1.0 and self.result.value > self.upper_threshold:
```
joker234, line 125: Just a suggestion:
```suggestion
        if 1.0 >= self.result.value > self.upper_threshold:
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +45,183 @@ def __init__(
...
-            # growth level is better than the red threshold
-            else:
-                self.result.label = "yellow"
-                self.result.value = 0.5
-                self.result.description = (
-                    description + self.metadata.label_description["yellow"]
+        elif (
+            self.result.value <= self.upper_threshold
+            and self.result.value > self.lower_threshold
+        ):
```
joker234, line 133:
```suggestion
        elif self.upper_threshold >= self.result.value > self.lower_threshold:
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
```diff
@@ -35,162 +45,183 @@ def __init__(
...
+            self.result.label = "yellow"
+            self.result.description = (
+                description + self.metadata.label_description["yellow"]
+            )
+        elif self.result.value <= self.lower_threshold and self.result.value > 0:
```
joker234, line 138: Just a suggestion:
```suggestion
        elif self.lower_threshold >= self.result.value > 0:
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/metadata.yaml
```diff
@@ -3,29 +3,15 @@ MappingSaturation:
   name: Mapping Saturation
   description: |
     Calculate if mapping has saturated.
+    High saturation has been reached if the growth of the fitted curve is minimal.
   label_description:
     red: |
-      Saturation has not been reached at all.
-      The mapping in this region are in just starting.
-      With regard to completeness this means bad quality.
+      No saturation identified (Saturation <= 30%).
```
joker234, line 9:
```suggestion
      No saturation identified (Saturation ≤ 30%).
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/metadata.yaml
```diff
@@ -3,29 +3,15 @@ MappingSaturation:
...
     yellow: |
-      Saturation not yet reached. The data in this region still seem to be in
-      a growth stadium with a growth of data more than 3 % within the last 3 years.
-      This indicates a difficult to specify quality in respect to completeness,
-      because it depends on the growth rate over time. The growth could be
-       near to saturated or still be in an high growth stadium.
+      Saturation is in progress (30% < Saturation <= 97%).
```
joker234, line 11:
```suggestion
      Saturation is in progress (30% < Saturation ≤ 97%).
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/metadata.yaml
```diff
@@ -3,29 +3,15 @@ MappingSaturation:
...
     green: |
-      Saturation has been reached. The data in this region seem quite saturated
-      with a growth of data less than 3 % within the last 3 years.
-      This indicates good quality in respect to completeness.
+      High saturation has been reached (97% < Saturation <= 100%).
```
joker234, line 13:
```suggestion
      High saturation has been reached (97% < Saturation ≤ 100%).
```

joker234 (changes requested):
Open for further review for @joker234:
- [x] workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
- [x] workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
- [ ] workers/ohsome_quality_analyst/indicators/mapping_saturation/ssdoubles.R
- [x] workers/scripts/run_mapping_saturation_models.py
- [ ] workers/tests/

CHANGELOG.md
```diff
@@ -6,14 +6,23 @@
...
 - Disable support of the parameter `bpolys` for GET requests to the API ([#223])
+- Rewrite of the Mapping Saturation indicator to use statistical models from the R language ([#170])
+
+### Other Changes
+
+- Use [`scipy`] to fit sigmoid curves to data for the Mapping Saturation indicator ([#170])
+
+### Other Changes
 
 ### How to upgrade
```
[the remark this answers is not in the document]
matthiasschaub: Added a single sentence to make sure R is available on the system when upgrading.

workers/ohsome_quality_analyst/api/api.py
matthiasschaub: Fixed

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
matthiasschaub: Just created one 
https://github.com/GIScience/ohsome-quality-analyst/issues/248

workers/ohsome_quality_analyst/indicators/mapping_saturation/indicator.py
matthiasschaub: True. Thats the place where it should be documented. 
I added some documentation to the base Result class

workers/setup.cfg
matthiasschaub: Good catch. Its not. I removed it.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - `ideal_vgi_lulc` is not a valid layer for the Mapping Saturation indicator anymore ([#221])
 - For requests to the `/indicator` or `/report` endpoints of the API for a custom AOI (usage of the `bpolys` parameter): use the POST method ([#223])
+- `R` (≥ 4.0) needs to be available on the system on which OQT runs ([#170t])
```
joker234, line 21:
```suggestion
- `R` (≥ 4.0) needs to be available on the system on which OQT runs ([#170])
```

workers/ohsome_quality_analyst/base/indicator.py
```diff
@@ -42,7 +42,17 @@ class LayerDefinition:
...
 class Result:
-    """The result of the Indicator."""
+    """The result of the Indicator.
+
+    Attributes:
+        timestamp_oqt (datetime): Timestamp of the creation of the indicator
+        timestamp_osm (datetime): Timestamp of the used OSM data
+            (e.g. Latest timestamp of the ohsome API results)
+        label (str): Traffic lights like quality label
+        value (float): The result value as float (0=>=1)
```
joker234, line 52: To me `=>=` is very uncommon. Do you have a better representation like `0 ≥ value ≥ 1` or `[0, 1]`, …
```suggestion
        value (float): The result value as float ([0, 1])
```

workers/ohsome_quality_analyst/utils/helper_dict.py
```diff
@@ -0,0 +1,138 @@
...
+        if prefix != "":
+            prefix += separator
+        for key, _ in input_.items():
+            output.update(
+                flatten_dict(
+                    input_[key],
+                    separator=separator,
+                    prefix=prefix + key,
+                ),
+            )
```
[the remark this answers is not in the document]
joker234: will be changed in #247

workers/ohsome_quality_analyst/utils/helper_dict.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Works only on first level of a nested dictionary.
+
+    Returns:
+        list|dict: List of values if dict keys are integers, input dictionary otherwise.
+    """
+    if all([is_int(k) for k in input_dict.keys()]):
+        return list(input_dict.values())
+    else:
+        return input_dict
```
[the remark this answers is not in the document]
joker234: will be changed in #247

workers/ohsome_quality_analyst/utils/helper_dict.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def is_int(string: str) -> bool:
+    """Check if string can be cast to integer."""
+    try:
+        int(string)
+        return True
+    except ValueError:
+        return False
```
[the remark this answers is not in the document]
joker234: will be changed in #247

workers/ohsome_quality_analyst/utils/helper_dict.py
```diff
@@ -0,0 +1,138 @@
...
+        input_seq = input_seq.values()
+    for val in input_seq:
+        if isinstance(val, (dict, list, tuple, set)):
+            output += flatten_sequence(val)
+        else:
+            output.append(val)
+    return output
+
+
+def flatten_dict(input_: dict, *, separator: str = ".", prefix: str = "") -> dict:
```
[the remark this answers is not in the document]
joker234: not necessary, since the lists are not used in the website, related to #247

workers/ohsome_quality_analyst/utils/helper_dict.py
```diff
@@ -0,0 +1,138 @@
...
+        keys = k.split(separator)
+        temp_dict = {}
+        for i, key in enumerate(reversed(keys)):
+            if i == 0:
+                temp_dict = {key: v}
+            else:
+                temp_dict = {key: temp_dict}
+        output_dict = merge_dicts(output_dict, temp_dict)
+    if reverse:
+        return dict_items_to_list(output_dict)
```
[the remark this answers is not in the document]
joker234: see #247

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+            "x_0": popt[0],
+            "k": popt[1],
+            "L": popt[2],
+        }
+        self.asym_conf_int = self.confint(popt, pcov)
+        self.fitted_values = self.function(xdata, **self.coefficients)
+
+    def function(
+        self, x: np.float64, x_0: np.float64, k: np.float64, L: np.float64
+    ) -> np.float64:
```
joker234, line 120: - [ ] `x` gets a `ndarray` not a `float64`. Please check what makes more sense.
- [ ] `python:S117`

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+        """Sigmoid function."""
+        return L / (1 + np.exp(-k * (x - x_0)))
+
+    def initial_guess(self) -> tuple:
+        """Get an initial guess on parameters for single Sigmoid function."""
+        x_0 = self.xdata.size / 2
+        k = 0
+        L = self.ydata.max()
+        return (x_0, k, L)
```
joker234, line 129:
```suggestion
        return x_0, k, L
```

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+        self, x: np.float64, x_0: np.float64, k: np.float64, L: np.float64
+    ) -> np.float64:
+        """Sigmoid function."""
+        return L / (1 + np.exp(-k * (x - x_0)))
+
+    def initial_guess(self) -> tuple:
+        """Get an initial guess on parameters for single Sigmoid function."""
+        x_0 = self.xdata.size / 2
+        k = 0
+        L = self.ydata.max()
```
joker234, line 128: - [ ] For `ndarray.max()`: set initial value
- [ ] capital `L`

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+        return (x_0, k, L)
+
+    def bounds(self) -> tuple:
+        """Get lower and upper bounds of the parameters for single Sigmoid function."""
+        x_0_upper_bound = self.xdata.size * 1.5
+        x_0_lower_bound = 0.0
+        k_upper_bound = 1.0
+        k_lower_bound = -1.0
+        L_upper_bound = self.ydata.max()
+        L_lower_bound = 0.0
```
joker234, line 138: `python:S117`

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+        L = self.ydata.max()
+        return (x_0, k, L)
+
+    def bounds(self) -> tuple:
+        """Get lower and upper bounds of the parameters for single Sigmoid function."""
+        x_0_upper_bound = self.xdata.size * 1.5
+        x_0_lower_bound = 0.0
+        k_upper_bound = 1.0
+        k_lower_bound = -1.0
+        L_upper_bound = self.ydata.max()
```
joker234, line 137: For `ndarray.max()`: set initial value

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+
+    name = "Two-Steps-Sigmoidal Model (Tangens Hyperbolicus)"
+    function_formula = (
+        "e + (f - e) * 1 / 2 * (np.tanh(k * (x - b)) + 1)"
+        + "+ (Z - f) * 1 / 2 * (np.tanh(k * (x - c)) + 1)"
+    )
+
+    def __init__(self, xdata, ydata):
+        super().__init__(xdata, ydata)
+        if xdata.min() == 0:
```
joker234, line 245: For `ndarray.min()`: set initial value

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+            xdata = xdata + 1
+        if ydata.min() == 0:
```
joker234, line 247: For `ndarray.min()`: set initial value

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+    https://rdrr.io/r/stats/SSmicmen.html
+    """
+
+    name = "Nls Michaelis-Menten Model"
+    function_formula = "Vm * x / (K + x)"
+
+    def __init__(self, xdata, ydata):
+        # Model fails when xdata or ydata includes zero
+        super().__init__(xdata, ydata)
+        if xdata.min() == 0:
```
joker234, line 407: For `ndarray.min()`: set initial value

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
```diff
@@ -0,0 +1,428 @@
...
+            xdata = xdata + 1
+        if ydata.min() == 0:
```
joker234, line 409: For `ndarray.min()`: set initial value

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+from rpy2.rinterface_lib.embedded import RRuntimeError
+
+import ohsome_quality_analyst.geodatabase.client as db_client
+import ohsome_quality_analyst.ohsome.client as ohsome_client
+from ohsome_quality_analyst.base.indicator import LayerDefinition
+from ohsome_quality_analyst.indicators.mapping_saturation import models
+from ohsome_quality_analyst.utils.definitions import get_layer_definition
+
+
+def plot(xdata, ydata, fitted_models):
```
joker234, line 20: Avoid shadowing variables from outer scope.

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+    plt.figure()
+    prev_ax = None
+    for i, model in enumerate(fitted_models):
```
joker234, line 23: Avoid shadowing variables from outer scope.

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+        ax.axhline(y=model.asymptote, color="pink", linestyle="--", label="Asymptote")
+        ax.set_title(model.name)
+        ax.legend()
+        ax.set(xlabel="time", ylabel="features")
+        ax.label_outer()
+        prev_ax = ax
+    plt.show()
+
+
+async def query_ohsome_api(features, layers) -> list:
```
joker234, line 43: Avoid shadowing variables from outer scope.

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+    time_range = "2008-01-01//P1M"
+    list_of_values = []
```
joker234, line 45: Avoid shadowing variables from outer scope.

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+        for layer in layers:
+            query_results = await ohsome_client.query(
+                layer=layer, bpolys=feature.geometry, time=time_range
+            )
+            list_of_values.append([item["value"] for item in query_results["result"]])
+    return list_of_values
+
+
+def get_layers(layer_names) -> list:
+    layers = []
```
joker234, line 56: Avoid shadowing variables from outer scope.

workers/scripts/run_mapping_saturation_models.py
```diff
@@ -0,0 +1,127 @@
...
+    for layer_name in layer_names:
+        layers.append(
+            from_dict(data_class=LayerDefinition, data=get_layer_definition(layer_name))
+        )
+    return layers
+
+
+async def get_features() -> list:
+    fids = await db_client.get_feature_ids("regions")
+    features = []
```
joker234, line 66: Avoid shadowing variables from outer scope.

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
```diff
@@ -0,0 +1,29 @@
+"""Collection of common error metrics as functions."""
+
+import numpy as np
+from numpy import float64, ndarray
+
+
+def me(actual: ndarray, predicted: ndarray) -> float64:
+    """Mean Error"""
+    return np.mean(np.subtract(actual, predicted))
```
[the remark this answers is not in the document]
matthiasschaub: Not sure what the actualy diagnostic here is.

np.mean returns a float. I checked manually as well as with the docs.
np.substracts works an arraylike objects.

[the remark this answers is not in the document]
joker234: > **Returns**    **m :** ***ndarray, see dtype parameter above***
>
> If out=None, returns a new array containing the mean values, otherwise a reference to the output array is returned.

Source: https://numpy.org/doc/stable/reference/generated/numpy.mean.html

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
```diff
@@ -0,0 +1,29 @@
+"""Collection of common error metrics as functions."""
+
+import numpy as np
+from numpy import float64, ndarray
+
+
+def me(actual: ndarray, predicted: ndarray) -> float64:
```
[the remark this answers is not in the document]
matthiasschaub: Of all those functions only `mae` ("Mean Absolute Error") is used at the moment.

I thought it would be nice to have those errors implemented for the future. In fact at first we have not been sure about the metric for model selection.

On the other hand `sklearn` and with it access to various metrics will be available for OQT soon. 

I will remove the other metrics.

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def nrmse(actual: ndarray, predicted: ndarray) -> float64:
+    """Normalized Root Mean Squared Error"""
+    return rmse(actual, predicted) / (actual.max() - actual.min())
+
+
+def mae(actual: ndarray, predicted: ndarray) -> float64:
+    """Mean Absolute Error"""
+    return np.mean(np.abs(np.subtract(actual, predicted)))
```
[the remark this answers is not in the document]
matthiasschaub: Remove metrics file since only one metric is used. Implement metric directly in base model class.
 https://github.com/GIScience/ohsome-quality-analyst/pull/170/commits/e8e5df0891672c643733abba7b39d18a03913515

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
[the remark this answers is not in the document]
matthiasschaub: Remove metrics file since only one metric is used. Implement metric directly in base model class.
 https://github.com/GIScience/ohsome-quality-analyst/pull/170/commits/e8e5df0891672c643733abba7b39d18a03913515

workers/scripts/run_mapping_saturation_models.py
matthiasschaub: This should be fixed now.

workers/scripts/run_mapping_saturation_models.py
matthiasschaub: This should be fixed now

workers/scripts/run_mapping_saturation_models.py
matthiasschaub: This should be fixed now

workers/scripts/run_mapping_saturation_models.py
matthiasschaub: This should be fixed now

workers/scripts/run_mapping_saturation_models.py
joker234: `models` is already defined, see line 16.

matthiasschaub: Fixed by renaming the variable.
I fixed all those issue by simply renaming the variables even though it may not be beautiful code. But this script is only used for an easy way to create all models for all regions and that only for test porpuses. So I did not want to spend much time on it.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: I would like to use variable names as they are used in the function definition. In this case capital `L`.
This is also setup to be ignored setup.py for flake8 and sonar conf for sonarcloud.
https://github.com/GIScience/ohsome-quality-analyst/blob/a8dd694b4909e89fdf3e101f2458aa341be92eaf/sonar-project.properties#L21

matthiasschaub: I set the initial values of all the min/max functions to zero.

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: Done

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: I would like to use variable names as they are used in the function definition. In this case capital L.
This is also setup to be ignored setup.py for flake8 and sonar conf for sonarcloud.
https://github.com/GIScience/ohsome-quality-analyst/blob/3c88e43dc9d35ef978083e0311edc7e8d3e6e5c1/sonar-project.properties#L21

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: I would like to use variable names as they are used in the function definition. In this case capital `L`.
This is also setup to be ignored setup.py for flake8 and sonar conf for sonarcloud.
https://github.com/GIScience/ohsome-quality-analyst/blob/a8dd694b4909e89fdf3e101f2458aa341be92eaf/sonar-project.properties#L21

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: Done

workers/ohsome_quality_analyst/indicators/mapping_saturation/models.py
matthiasschaub: Done

workers/scripts/run_mapping_saturation_models.py
matthiasschaub: Done

joker234 (approved):

joker234 (approved):

PR #209, opened 2021-11-24
Mapping saturation sigmoid curve rpy2

### LINKED ISSUE #208: Implement additional regression models for the Mapping Saturation indicator
Implement additional regression models (Functions for fitting optimal curves) for the Mapping Saturation indicator. (Details on what regression models will follow in the PR)

Three regression models should be based on following R models:

- SSlogis
- SSfpl
- SSasymp

Those three are self-starting models in R. That means the initial guess are taking care of by the model itself. To use these models in Python following possibilities are present:

1. Implement those models in Python: Meaning use the same formulae, make an initial guess and use scipy.curve_fit to find best fit. This is done in the current implementation for a simple sigmoid function.
2. Find similar self starting model in third party Python libraries
3. Execute R code from Python.

# 3. Execute R code from Python

```python
from rpy2.robjects import FloatVector, globalenv, r

x = [1, 2, 3, 4, 5, 6, 7]
y = [5, 5, 80, 100, 500, 500, 500]

globalenv["x"] = FloatVector(x)
globalenv["y"] = FloatVector(y)
coef = r(
    """
df <- data.frame(x, y)
fm <- nls(y ~ SSlogis(x, Asym, xmid, scal), data = df)
coef(fm)
"""
)
print(coef)
print(coef[0])  # Asym
print(coef[1])  # xmid
print(coef[2])  # scal
```

matthiasschaub:
### Description

Implement additional regression models (Functions for fitting optimal curves) for the Mapping Saturation indicator. The usage of those models are based on current research. R and its stats module is used to run those models. Communitcation between Python and R is done by the Python package `rpy2`.

The 5 additional models implemented are:
- Self-Starting Nls Michaelis-Menten Model: `Vm * x / (K + x)`
- Self-Starting Nls Asymptotic Regression Model: `asym + (R0 - asym) * e^(-e^(lrc) * x)`
- Self-Starting Nls Four-Parameter Logistic Model: `A + (B - A) / (1 + e^((xmid - x) / scal))`
- Self-Starting Nls Logistic Model: `asym / (1 + e^((xmid - x) / scal))`
- Two-Steps-Sigmoidal / Tangens Hyperbolicus: (See below)
```
e + (f - e) * 1 / 2 * (np.tanh(k * (x - b)) + 1) + (Z - f) * 1 / 2 * (np.tanh(k * (x - c)) + 1)
```

### Corresponding issue
Closes #208 

- System Package: `R`
- Python: `rpy2`

### Checklist

### Extra Checklist
- [x] Go over TODO comments in code
- [x] Fix failure of Michaelis-Menten model due to input values containing zero.
- [x] Run all models for a bunch of input data and plot those
- [x] Decide whether to refactor `fit.py` into `indicator.py` and `model.py`
- [x] Add R as depedency to docs and Dockerfile
- [x] Update PR description with more details
- [x] Decide on with sigmoidal models to keep (SSlogis and Tangens Hyperbolicus vs previous implemented Sigmoid and DoubleSigmoid model)
- [x] Should fit with `nls` be weighted?
  - No, but mark possibillity in code


matthiasschaub:
Figure_1

workers/scripts/mapping_saturation.py
```diff
@@ -0,0 +1,57 @@
...
+from tests.unittests.mapping_saturation.fixtures import VALUES
+
+XDATA = np.asarray(range(len(VALUES)))
+YDATA = VALUES
+
+
+def run():
+    fits = run_all_models(XDATA, YDATA)
+
+    fig, axs = plt.subplots(2, 3, sharex=True, sharey=True)
```
Hagellach37, line 16: I often use something like this:
```
import matplotlib.pyplot as plt
import matplotlib.gridspec as gridspec

fig = plt.figure(figsize=(12.5, 6))
gs1 = gridspec.GridSpec(2, 3)
gs1.update(wspace=0.2, hspace=0.35) # set the spacing between axes. 

for i in range(0, len(fits)):
    ax = plt.subplot(gs1[i])
```

the you can just loop through the subplots and get the axes

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
```diff
@@ -1,3 +1,5 @@
+"""Collection of common error metrics as functions."""
+
```
Hagellach37, line 2: Just out of curiosity, why are we not relying on the functions provided by scikit-learn here? Because we want to avoid this addtional dependy?

Another question on the `mean error`: usually one would use the mean absolute error to evaluate the model performance.

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -1,111 +1,59 @@
...
+        models.Sigmoid(),
+        models.SSlogis(),
+        models.SSdoubleS(),
+        models.SSfpl(),
+        models.SSasymp(),
+        # models.SSmicemen(),
+    ):
+        coef = model.fit(xdata, ydata)
+        ydata_fitted = model.function(xdata, **coef)
+        metric = metrics.mae(ydata, ydata_fitted)
```
Hagellach37, line 36: Does this metric exist? In the metrics you just have `me`.

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -1,111 +1,59 @@
...
+        fit = Fit(
```
Hagellach37, line 37: I'm a bit confused by these lines. Maybe you can help me and explain what you are doing here. I'm not sure what you "fit" here as you already have `ydata_fitted` a few lines above. 

From my understanding you did the curve fitting already, but here you try to estimate the model performance of your fitted values.

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
matthiasschaub: Yes, thats right. No need for an additional dependency since the calculation of those error metrics are pretty straightforward. Have a closer look at this file (`metrics.py`). The `mean error` is just one of the implemented errors. the `mean absolute error` is also implemented and currently used as criterion to choose the best model.

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
matthiasschaub: Regarding the metrics: Have a closer look at the `metrics.py` file. In this file multiple error metrics are implemented. Its just that the diff is not showing the difference since this PR is not directed at the main branch.

matthiasschaub: I see the reason of the confusion here. The variable names are not very precise.
I renamed the dataclass `Fit` to `FittedModel`. To propose of the dataclass is to collect all relevant data of the model result. So there is no fitting going on in this line, just to collection of the result (fitted values, metric, etc.).

In fact this file (`fit.py`) could probably be refactored into either `models.py` or `indicator.py` (or into both). E.g. all attributes of the `FittedModel` dataclass could actually be attributes of the model class itself.

workers/scripts/mapping_saturation.py
matthiasschaub: Thats a nice solution. Thanks for the hint. 
I choose a different solution: Now the script creates subplots on the fly.

workers/ohsome_quality_analyst/indicators/mapping_saturation/metrics.py
Hagellach37: ah perfect, yeah I missed the last lines of this file.

workers/ohsome_quality_analyst/indicators/mapping_saturation/fit.py
```diff
@@ -1,111 +1,71 @@
...
+        models.SSlogis(),
+        models.SSdoubleS(),
+        models.SSfpl(),
+        models.SSasymp(),
+        models.SSmicmen(),
+    ):
+        logging.info("Run {}".format(model.name))
+        try:
+            coef = model.fit(xdata, ydata)
+        except RRuntimeError as error:
```
Hagellach37, line 39: you might need to change this to `except (RRuntimeError, RuntimeError) as error:`

I tested it locally and got an error that was initially not catched by your line.

```
2021-12-09 17:40:18,791 - WARNING - callbacks.py - consolewrite_warnerror - R[write to console]: Error in nls(y ~ cbind(1, 1/(1 + exp((xmid - x)/exp(lscal)))), data = xy,  : 
  number of iterations exceeded maximum of 50

2021-12-09 17:40:18,792 - WARNING - fit.py - run_all_models - Could not run model Self-Starting Nls Four-Parameter Logistic Model due to RRuntimeError: Error in nls(y ~ cbind(1, 1/(1 + exp((xmid - x)/exp(lscal)))), data = xy,  : 
  number of iterations exceeded maximum of 50
```

Hagellach37: `Could not run model Self-Starting Nls Four-Parameter Logistic Model due to RRuntimeError` is logged by you manually.

matthiasschaub: The first of the two error message which you shared above is actually just a logging message. Before the `RRuntimeError` is raised the rpy2 library will log the message and then raise the error. The first of the two logging messages above is from rpy2 then the error gets raised which is caught by OQT and error messages is logged again, resulting in two logs with the same message. Does this make sense?

PR #221, opened 2022-01-03
Remove layer definition "ideal_vgi_lulc"

matthiasschaub:
### Description
Layer definition "IDEAL-VGI Land Use and Land Cover" has been a valid
layer in combination with the Mapping Saturation indicator.
But often time the area of the land-use gets smaller over time instead
of bigger, which makes it not suitable for the Mapping Saturation.
Beside that the Layer has not been used in any Report.

### Checklist
~- [ ] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#pre-commit) before committing~
~- [ ] I have commented my code~
~- [ ] I have added sufficient unit and integration [tests](https://github.com/GIScience/ohsome-quality-analyst/blob/main/docs/development_setup.md#tests)~
- [x] Decide on whether below mentioned alternatives would be better the solution proposed in this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+### Breaking Changes
+
+- Remove "IDEAL-VGI Land Use and Land Cover" as valid layer for the Mapping Saturation indicator ([#221])
+
+### How to upgrade
+
+- `ideal_vgi_lulc` is not a valid layer for the Mapping Saturation indicator anymore ([#221])
+
+[#221]: https://github.com/GIScience/ohsome-quality-analyst/pull/221
+
+
 ## 0.7.0
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


matthiasschaub:
Two alternatives come to mind. Both would keep the layer definition and ...
1. Remove layer definition as valid layer for the Mapping Saturation indicator
2. Keep as it was and add checks to the Mapping Saturation indicator

Hagellach37:
Hey @matthiasschaub ,
I think from my perspective it might be enough to remove this from the `definitions.py` file and keep in the layer definitions.

Hagellach37 (dismissed):

PR #255, opened 2022-02-21
Configure OQT using files or environment variables

matthiasschaub:
### Description
Configure OQT using files or environment variables.

All configuration can be done using environment variables.

All configuration can also be done by providing a configuration file (`yaml` format). 
The default configuration file path is `workers/config/config.yaml`. The path to the configuration can be changed by setting the path in the environment variable `OQT_CONFIG`. 
A sample configuration file is provided at `workers/config/config.yaml`.

### Corresponding issue
- None

### Checklist
- [x] Update import statements of definitions.py
- [x] Write configuration docs
    - [x] Provide a table with all possible configuration variables
    - [x] Provide prose on how to setup config files
- [x] Clean up commit history
- [x] Add configuration for semaphore (#307)

Open question:
- Should configuration variables be made available as globals through definitions.py? -> No. Two functions to get either all config vars or a single value for a given key are provided.
- Tests for config are very minmal. Should more time be spent to implement further tests?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Breaking Changes
 
+- Rename environment variable `OHSOME_API` to `OQT_OHSOME_API` ([#255])
+
+### Bug Fixes
+
 - Make inclusion of indicator data in response optional ([#370])
 - Per default properties of the GeoJSON response are not flat ([#375])
 - Remove project specific reports `JrcRequirements`, `SketchmapFitness`, and `MapActionPoc` from the website ([#382])
@@ -15,6 +19,7 @@
 - Add new representative report `RoadReport` ([#357])
 - Add new representative report `BuildingReport` ([#356])
 - Add ratio_filter to `building_count` layer ([#356])
+- Configure OQT using files or environment variables ([#255])
 
 ### Other Changes
 
@@ -28,6 +33,7 @@
 - Reports `JrcRequirements`, `SketchmapFitness`, and `MapActionPoc` are not accessibly via the website anymore. If you want to access those reports please use the API.
 - To continue to retrieve the properties of the GeoJSON API response as flat list, you need to set the API request parameter `flatten` to `True` ([#375])
 - To continue to retrieve additional data of an Indicator or Report provided in an API response, you need to set the API request parameter `include_data` to `True` ([#370])
+- Rename environment variable `OHSOME_API`  `OQT_OHSOME_API` ([#255])
 - Make sure to rename the API query parameter `layerName` to `layerKey` and API endpoint `listLayerNames` to `listLayerKeys` ([#376])
 - To continue to retrieve the properties of the GeoJSON API response as flat list, you need to set the API request parameter `flattem` to `True` ([#375])
 - Rename endpoints ([#397]):
@@ -40,6 +46,7 @@
 | `reportNames`                | `reports`                      |
 | `fidFields`                  | `fid-fields`                   |
 
+[#255]: https://github.com/GIScience/ohsome-quality-analyst/pull/255
 [#342]: https://github.com/GIScience/ohsome-quality-analyst/pull/342
 [#356]: https://github.com/GIScience/ohsome-quality-analyst/pull/356
 [#357]: https://github.com/GIScience/ohsome-quality-analyst/pull/357
@@ -77,6 +84,7 @@
 - Add `flatten` parameter to API request. Make flatten of GeoJSON properties of Indicators and Reports optional ([#303])
 - Make calculation of an Indicator for a FeatureCollection or for a Report asynchronous ([#307])
 - Add new Indicator which predicts the building area of the AOI using a trained Random Forest Regressor ([#265])
+- Add support for `groupBy/boundary` queries to the ohsome API client ([#272])
 
 ### Other Changes
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,48 @@
+# Configuration
+
+OQT can be configured using a configuration file or environment variables. Configuration
+values from environment variables have precedence over values from the configuration
+file.
+
+Below is a table listing all possible configuration variables.
+
+| Configuration Variable Name | Environment Variable Name     | Configuration File Name   | Default Value                                       | Description                                                                  |
+| --------------------------- | -------------------------     | -----------------------   | --------------------------------------------------- | ---------------------------------------------------------------------------- |
+| Postgres Host               | `POSTGRES_HOST`               | `postgres_host`           | `localhost`                                         | Database connection parameter                                                |
+| Postgres Port               | `POSTGRES_PORT`               | `postgres_port`           | `5445`                                              | "                                                                            |
+| Postgres Database           | `POSTGRES_DB`                 | `postgres_db`             | `oqt`                                               | "                                                                            |
+| Postgres User               | `POSTGRES_USER`               | `postgres_user`           | `oqt`                                               | "                                                                            |
+| Postgres Password           | `POSTGRES_PASSWORD`           | `postgres_password`       | `oqt`                                               | "                                                                            |
+| Configuration File Path     | `OQT_CONFIG`                  | -                         | `workers/ohsome_quality_analyst/config/config.yaml` | Absolute path to the configuration file                                      |
+| Data Directory              | `OQT_DATA_DIR`                | `data_dir`                | `data/`                                             | Absolute path to the directory for raster files                              |
+| Datasets and Features IDs   | -                             | `datasets`                | `[{"regions": {"default": "ogc_fid"}}]`             | Dataset and Features Ids available in the database (see description below)   |
+| Geometry Size Limit (km²)   | `OQT_GEOM_SIZE_LIMIT`         | `geom_size_limit`         | `100`                                               | Area restriction of the input geometry to the OQT API (sqkm)                 |
+| Python Log Level            | `OQT_LOG_LEVEL`               | `log_level`               | `INFO`                                              | Python logging level                                                         |
+| Concurrent Computations     | `OQT_CONCURRENT_COMPUTATIONS` | `concurrent_computations` | `4`                                                 | Limit number of concurrent Indicator computations for one API request        |
+| User Agent                  | `OQT_USER_AGENT`              | `user_agent`              | `ohsome-quality-analyst/{version}`                  | User-Agent header for requests tot the ohsome API                            |
+| ohsome API URL              | `OQT_OHSOME_API`              | `ohsome_api`              | `https://api.ohsome.org/v1/`                        | ohsome API URL                                                               |
+
+_Note on the 'Datasets and Features IDs' configuration:_ Defines the datasets and
+features IDs which are available in the database and for which Indicators should be
+calculated. The `dataset` field refers to the name of the relation and `default` and
+`other` refer to possible identifier of the features. The default configuration looks
+like this:
+
+```yaml
+datasets:
+    regions:
+        default: ogc_fid
+        other: [name] # Optional
+```
+
+## Configuration File
+
+The default path of the configuration file is `workers/config/config.yaml`.
+A sample configuration file can be found in the same directory: `workers/config/sample.config.yaml`.
+All configuration files in this directory (`config`) will be ignored by Git. To change the default configuration file path for OQT set the environment variable `OQT_CONFIG` to the desired path.
+
+To create a new configuration file simply copy the sample configuration file and change the values.
+
+```
+cp sample.config.yaml config.yaml
+```
--- a/docs/development_setup.md
+++ b/docs/development_setup.md
@@ -79,40 +79,9 @@ pre-commit install  # Install pre-commit hooks.
 
 ### Configuration
 
-
-#### Local database
+For all possible configuration parameter please refer to the [configuration documentation](/docs/configuration.md).
 
 For local development no additional configuration is required. Per default OQT will connect to the database defined in `docker-compose.development.yml`.
-
-
-#### Remote database
-
-If access to a remote database is required following environment variables need to be set:
-
-```bash
-POSTGRES_DB
-POSTGRES_USER
-POSTGRES_PASSWORD
-POSTGRES_HOST
-POSTGRES_PORT
-POSTGRES_SCHEMA
-```
-
-> Tip: Above lines can be written to a file (e.g. `.env`), prefixed with `export` and sourced (`source .env`) to make them available to current environment.
->
-> Note: Windows user can set those environment variables with following command `setx POSTGRES_DB`
-
-
-#### ohsome API
-
-The URL to a specific ohsome API can be set with the environment variable `OHSOME_API`. It defaults to [https://api.ohsome.org/v1/](https://api.ohsome.org/v1/)
-
-
-#### Additional options
-
-Additional environment variables are:
-- `OQT_LOG_LEVEL`: Control the logging level of OQT (See logging section)
-- `OQT_GEOM_SIZE_LIMIT`: Control the size limit of the input geometry passed to the API.
 
 
 ### Usage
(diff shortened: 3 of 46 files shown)


workers/tests/unittests/test_cli.py
```diff
@@ -57,6 +57,9 @@ def test_create_all_indicators_invlid_opts(self, mo):
         assert result.exit_code == 2
 
     @mock.patch("ohsome_quality_analyst.oqt.create_indicator", mock.AsyncMock())
+    @mock.patch(
+        "ohsome_quality_analyst.geodatabase.client.get_feature_ids", mock.AsyncMock()
+    )
```
joker234, line 62: What was changed to make this mock patch necessary?

workers/tests/unittests/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_get_config_path_empty_env(self):
+        self.assertEqual(
+            config.get_config_path(),
+            os.path.join(
+                os.path.dirname(
+                    inspect.getfile(config),
+                ),
+                "config.yaml",
+            ),
+        )
```
joker234, line 21: I would write this test differently and not copy the definition of the method into the unittest. I would compare it explicitly with the path string you are using.

workers/tests/unittests/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @mock.patch.dict(
+        "os.environ",
+        {"OQT_CONFIG": os.path.join(tempfile.gettempdir(), "config.yaml")},
+        clear=True,
+    )
+    def test_get_config_path_set_env(self):
+        self.assertEqual(
+            config.get_config_path(),
+            os.path.join(tempfile.gettempdir(), "config.yaml"),
```
joker234, line 31: see above. It shouldn't be *exactly* the same implementation. I would use strings here.

workers/tests/unittests/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+            {"geom_size_limit": "200", "postgres_host": "foo"},
+        )
+
+    @mock.patch.dict("os.environ", {}, clear=True)
+    def test_get_config(self):
+        cfg = config.get_config()
+        self.assertIsInstance(cfg, dict)
+        self.assertTrue(cfg)
+
+    def test_check_missing_conf_keys_compete(self):
```
joker234, line 63:
```suggestion
    def test_check_missing_conf_keys_complete(self):
```

workers/ohsome_quality_analyst/config/config.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def check_missing_conf_keys(cfg) -> dict:
+    """Check for missing configuration keys.
+
+    Returns:
+        dict: Input dict without changes
+    Raises:
+        KeyError: In case configuration key is missing from configuration dictionary.
+    """
```
joker234, line 89: What is the semantic difference between a missing key and a key explicitly set to `None`/`none`? I would say there isn't any and therefore `None`-keys should be ignored.

workers/tests/unittests/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+            "postgres_user": None,
+            "postgres_password": None,
+            "data_dir": None,
+            "geom_size_limit": None,
+            "log_level": None,
+            "ohsome_api": None,
+            "test_db": None,
+            "user_agent": None,
+            "datasets": None,
+        }
```
joker234, line 77: I think `None` should be replaced through some values. See comment to `config.check_missing_conf_keys(…)`.

workers/tests/unittests/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+        expected = os.path.join(
+            os.path.dirname(
+                os.path.abspath(__file__),
+            ),
+            "..",
+            "..",
+            "..",
+            "data",
+        )
+        self.assertTrue(os.path.samefile(data_dir, expected))
```
joker234, line 97: This tests the same implementation in the test and the method. I would suggest to use a string instead.

workers/tests/unittests/test_definitions.py
```diff
@@ -73,41 +54,6 @@ def test_get_raster_dataset_undefined(self):
...
-        with tempfile.TemporaryDirectory() as tmpdirname:
-            custom_data_dir = os.path.join(tmpdirname, "oqt-custom")
-            os.mkdir(custom_data_dir)
-            with mock.patch.dict(
-                "os.environ",
-                {"OQT_DATA_DIR": custom_data_dir},
-                clear=True,
-            ):
-                data_dir = definitions.get_data_dir()
-            self.assertEqual(data_dir, custom_data_dir)
```
joker234, line 100: I couldn't really find a replacement for this test. Is this not necessary anymore?

workers/tests/unittests/test_raster.py
```diff
@@ -20,23 +22,20 @@ def setUp(self):
...
+        ).touch()
+
         self.assertEqual(
             raster_client.get_raster_path(self.raster_dataset),
             os.path.join(
-                "test_dir", "GHS_BUILT_LDS2014_GLOBE_R2018A_54009_1K_V2_0.tif"
+                tempfile.gettempdir(),
+                "GHS_BUILT_LDS2014_GLOBE_R2018A_54009_1K_V2_0.tif",
             ),
         )
```
joker234, line 37: Tests shouldn't mess with files in my `/tmp`. Please always create a tempdir/-file and remove it after the tests are done.

joker234 (changes requested):
Thanks for the PR. I have made some comments/questions, mostly about the details of the tests and the config design.

Two general remarks:
1. Is there any config variable without a default and mandatory to be set by the user? If no, I would really prefer to not having the config file mandatory.
2. The tests cannot depend on the local configuration. That doesn't make the tests independent, which they should be.

workers/ohsome_quality_analyst/config/sample.config.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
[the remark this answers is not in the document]
matthiasschaub: Even if documentation about those exists in the docs directory? There are nor mandatory options.

workers/ohsome_quality_analyst/config/sample.config.yaml
```diff
@@ -0,0 +1,20 @@
...
+postgres_password: oqt
+# Default is the root of the repo
+data_dir: null
+# Restrict size of input geometry to the OQT API (sqkm)
+geom_size_limit: 100
+log_level: INFO
+ohsome_api: https://api.ohsome.org/v1/
+test_db: False
+# Default is 'ohsome-quality-analyst/{version}'
+user_agent: null
```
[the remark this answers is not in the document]
matthiasschaub: No more `null` anymore!

workers/Dockerfile
```diff
@@ -25,4 +25,5 @@ RUN python -m poetry install --no-ansi --no-interaction --no-root
 
 # copy all the other files and install the project
 COPY --chown=oqt:oqt . .
+RUN cp ohsome_quality_analyst/config/sample.config.yaml ohsome_quality_analyst/config/config.yaml
```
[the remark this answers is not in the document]
matthiasschaub: It is not anymore. I added a function to get default values for every config key. I reverted this change.

workers/ohsome_quality_analyst/config/config.py
matthiasschaub: This validation function has been removed in favor of default values if config is not provided.

workers/tests/unittests/test_config.py
matthiasschaub: Configuration setup has been changed to always use default values if no configuration has been provided

workers/ohsome_quality_analyst/config/config.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def load_config_from_file(path: str) -> dict:
+    """Load configuration from file on disk.
+
+    Remove items where value is None (`null` in YAML)
+    """
+    with open(path, "r") as f:
+        cfg = yaml.safe_load(f)
+    return {k: v for k, v in cfg.items() if v is not None}
```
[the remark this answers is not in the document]
matthiasschaub: Removed the sentence in docstring hinting that this is the case.

workers/ohsome_quality_analyst/config/sample.config.yaml
```diff
@@ -0,0 +1,20 @@
+---
+postgres_host: localhost
+postgres_port: 5445
+postgres_db: oqt
+postgres_user: oqt
+postgres_password: oqt
+# Default is the root of the repo
+data_dir: null
```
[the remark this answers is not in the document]
matthiasschaub: No more `null` values present anymore

workers/tests/unittests/test_config.py
matthiasschaub: This test has been removed since the function to be tested has been removed.

workers/ohsome_quality_analyst/definitions.py
```diff
@@ -50,6 +28,15 @@ class RasterDataset:
     crs: str
 
 
+CONFIG = get_config()
+# Dataset names and fid fields which are available in the Geodatabase
+DATASETS = MappingProxyType(CONFIG["datasets"])
+DATA_DIR = CONFIG["data_dir"]
+GEOM_SIZE_LIMIT = CONFIG["geom_size_limit"]
+OHSOME_API = CONFIG["ohsome_api"].rstrip("/")
+USER_AGENT = CONFIG["user_agent"]
```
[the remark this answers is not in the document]
matthiasschaub: There are not really necessary. I removed them and replaced their usages by accessing keys in `CONFIG` directly.

workers/tests/unittests/test_cli.py
matthiasschaub: Good catch. Nothing changed to make it necessary. I removed this mock.

workers/tests/unittests/test_config.py
matthiasschaub: The expected path is now build relative to the test file.

workers/tests/unittests/test_config.py
matthiasschaub: But then it is only valid for linux systems is it not.

workers/tests/unittests/test_raster.py
matthiasschaub: Now a tmp dir get created for this test

matthiasschaub:
> Thanks for the PR. I have made some comments/questions, mostly about the details of the tests and the config design.
> 
> Two general remarks:
> 
>     1. Is there any config variable without a default and mandatory to be set by the user? If no, I would really prefer to not having the config file mandatory.
> 
>     2. The tests cannot depend on the local configuration. That doesn't make the tests independent, which they should be.

Thanks for the throughout review so far!
After I addressed your review comments following statements about your general points are true:
1. There is now mandatory config variable to be set and a config file is not mandatory
2. The tests are not depending on a present config file anymore.

docs/development_setup.md
```diff
@@ -80,30 +80,11 @@ pre-commit install  # Install pre-commit hooks.
 
 ### Configuration
 
-
-#### Local database
+For all possible configuration parameter please refer to the [configuration documentation](/docs/configuration.md).
```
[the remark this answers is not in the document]
joker234: Why not? The configuration shouldn't have anything to do with the image but only being necessary when starting a container.

workers/tests/unittests/test_config.py
joker234: Okay maybe I just don't get why you're using the temp directory and not just any string/path?

workers/tests/unittests/test_config.py
```diff
@@ -31,15 +51,34 @@ def test_get_config_path_set_env(self):
             os.path.join(tempfile.gettempdir(), "config.yaml"),
         )
 
+    @mock.patch.dict("os.environ", {}, clear=True)
+    def test_config_default(self):
+        cfg = config.load_config_default()
+        self.assertIsInstance(cfg, dict)
+        self.assertListEqual(self.keys, list(cfg.keys()))
```
joker234, line 58: This comment applies to future occurrences as well:
Is the order of the keys relevant? `assertListEqual` checks the lists, which are ordered.

joker234 (commented):
**Changelog missing!**

workers/ohsome_quality_analyst/config/config.py
```diff
@@ -49,62 +73,30 @@ def load_config_from_env() -> dict:
...
-        "user_agent",
-        "datasets",
-    }
-    diff = config_keys.difference(set(cfg.keys()))
-    if diff:
-        raise KeyError("Configuration key(s) is/are missing: " + ", ".join(diff))
-    return cfg
+def get_config_value(key: str) -> Union[str, int, dict]:
+    config = get_config()
+    return config[key]
```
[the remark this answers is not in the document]
matthiasschaub: No not really. This error should only occur during development and that not often and easily fixable.

workers/ohsome_quality_analyst/definitions.py
```diff
@@ -123,15 +100,6 @@ class RasterDataset:
     ("TagsRatio", "jrc_bridge_count"),
     ("TagsRatio", "jrc_mass_gathering_sites_count"),
 )
-OHSOME_API = os.getenv("OHSOME_API", default="https://api.ohsome.org/v1/").rstrip("/")
```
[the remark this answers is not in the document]
matthiasschaub: Fixed. This is now reflected in both docker-compose files

workers/tests/unittests/test_config.py
matthiasschaub: Order is not necessary. Replaced with just `assertEqual`.

docs/configuration.md
```diff
@@ -1,27 +1,32 @@
...
+| Data Directory              | `OQT_DATA_DIR`            | `data_dir`              | `data/`                                             | Absolute path to the directory for raster files                              |
+| Datasets and Features IDs   | -                         | `datasets`              | `[{"regions": {"default": "ogc_fid"}}]`             | Dataset and Features Ids available in the database (see description below)   |
+| Geometry Size Limit (km²)   | `OQT_GEOM_SIZE_LIMIT`     | `geom_size_limit`       | `100`                                               | Area restriction of the input geometry to the OQT API (sqkm)                 |
+| Python Log Level            | `OQT_LOG_LEVEL`           | `log_level`             | `INFO`                                              | Python logging level                                                         |
+| Semaphore                   | `OQT_SEMAPHORE`           | `semaphore`             | `4`                                                 | Limit number of concurrent Indicator computations for one API request        |
```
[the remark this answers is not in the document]
matthiasschaub: Its not really just the limitation on concurrent request to the ohsome API, but a limitation on how many indicators are computed concurrently. Of course the requests to the ohsome API is where concurrency plays the biggest role.

`ohsome_api_limit` does not give hints about how the limit is archived (e.g. semaphore). I prefer just semaphore. This makes it clear that asynchronicity is at play and configurable.

CHANGELOG.md
```diff
@@ -2,23 +2,33 @@
...
+- Rename environment variable `OHSOME_API` to `OQT_OHSOME_API` ([#255])
+
 ### Bug Fixes
 
 - Add missing error handling of `RasterDatasetUndefinedError` and `RasterDatasetNotFoundError` to the API ([#298])
 
 ### New Features
 
 -  Add support for `groupBy/boundary` queries to the ohsome API client ([#272])
+- Configure OQT using files or environment variables ([#255])
```
joker234, line 16:
```suggestion
- Add support for `groupBy/boundary` queries to the ohsome API client ([#272])
- Configure OQT using files or environment variables ([#255])
```

docs/configuration.md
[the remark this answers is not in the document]
joker234: Something with `concurrency` would be nice.

workers/tests/unittests/test_config.py
joker234: Which doesn't change my question, since:

> These methods are used by default when comparing lists or tuples with [assertEqual()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual).

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual

matthiasschaub: Now both lists to be compared are sorted beforehand.

joker234 (dismissed):

joker234 (approved):

PR #362, opened 2022-06-22
set SHDI values for cells outside of intersection

matthiasschaub:
### Description
set SHDI values for cells outside of intersection to the mean of the SHDI value of all other cells.

### Corresponding issue
None

- None

### Checklist
- ~[ ] I have updated the [CHANGELOG.md](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CHANGELOG.md)~

### DIFF
--- a/workers/ohsome_quality_analyst/geodatabase/select_shdi.sql
+++ b/workers/ohsome_quality_analyst/geodatabase/select_shdi.sql
@@ -9,7 +9,8 @@ WITH bpoly AS (
 )
 SELECT
     SUM(ST_Area (ST_Intersection (shdi.geom, bpoly.geom)::geography) / ST_Area
-	(bpoly.geom::geography) * shdi.shdi) AS shdi
+	(bpoly.geom::geography) * shdi.shdi) AS shdi,
+    rownumber as rownumber
 FROM
     shdi,
     bpoly
--- a/workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
+++ b/workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
@@ -98,7 +98,6 @@ class BuildingCompleteness(BaseIndicator):
             item["result"][0]["value"] for item in query_results["groupByResult"]
         ]
         # Get covariates (input parameters or X)
-        shdi = await db_client.get_shdi(hex_cells)
         ghs_pop = raster_client.get_zonal_stats(
             hex_cells,
             get_raster_dataset("GHS_POP_R2019A"),
@@ -109,7 +108,7 @@ class BuildingCompleteness(BaseIndicator):
             get_raster_dataset("VNL"),
             stats=["sum"],
         )
-        self.covariates["shdi"] = [i["shdi"] for i in shdi]
+        self.covariates["shdi"] = await get_shdi(hex_cells)
         self.covariates["vnl"] = [i["sum"] for i in vnl]
         self.covariates["ghs_pop"] = [i["sum"] or 0 for i in ghs_pop]
         self.covariates["ghs_pop_density"] = [
@@ -203,7 +202,7 @@ class BuildingCompleteness(BaseIndicator):
             label="Weighted Average: {0}%".format(int(self.result.value * 100)),
         )
         ax.legend(loc="lower center", bbox_to_anchor=(0.5, -0.45))
-        # has to be excuted after "major formatter setting"
+        # has to be executed after "major formatter setting"
         plt.xlim(0, 1)
         plt.ylim(0, 10)
         img_data = StringIO()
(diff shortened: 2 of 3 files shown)


workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
```diff
@@ -98,7 +98,11 @@ async def preprocess(self) -> None:
             item["result"][0]["value"] for item in query_results["groupByResult"]
         ]
         # Get covariates (input parameters or X)
-        shdi = await db_client.get_shdi(hex_cells)
+        # Set default SHDI val to 0 for cells not intersecting with SHDI geometry
+        shdi = [{"sum": 0} for feature in hex_cells.features]
```
Hagellach37, line 102: do we use `sum` for SHDI, I thought it would be `mean` or something similar?

The default value should not be 0, but something more realistic? The minimum SHDI for a country is about 0.3 or 0.4

matthiasschaub: Renamed the result field back to just `shdi`

workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
```diff
@@ -271,3 +270,17 @@ async def get_hex_cells(feature: Feature) -> FeatureCollection:
     if feature_collection["features"] is None:
         raise HexCellsNotFoundError
     return feature_collection
+
+
+async def get_shdi(featurecollection: FeatureCollection) -> list:
```
Hagellach37, line 275: it's probably very unlikely, but what if we just get single hex cell without SHDI value? then you would not be able to calculate the mean.

workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
```diff
@@ -271,3 +270,17 @@ async def get_hex_cells(feature: Feature) -> FeatureCollection:
...
+    """Get SHDI values for each feature.
+
+    If feature does not intersect with the SHDI geometry take the mean of all present
+    SHDI values.
+    """
+    records = await db_client.get_shdi(featurecollection)
+    mean = np.mean([r["shdi"] for r in records])
+    shdi = [mean for feature in featurecollection.features]  # Set default
+    for r in records:
```
Hagellach37, line 284: to be honest, I do not really understand what you do here or why you do it this way. I think you do the following:

- get shdi values for hexgrids from database
- calculate mean
- create a new list with the length of all records and the mean shdi as value
- update the list with the actual  shdi value.

What I do not understand:
- how do you solve the problem that an shdi could be `null`
- why do you not handle this in the sql query directly? --> e.g. by setting a global default value for shdi

Hagellach37 (commented):
Hey @matthiasschaub ,

sorry that this discussion continues and seems to be longer than expected. Maybe you can take a look at my comments.

I'm not sure, but from my side we could either go for a very simple solution (e.g. use a fixed SHDI value we define as default for all hexagons regardless where they are located) or go with the distance based approach to assign a shdi value to a hexcell.

I think that it might be simple enough and good to go with the distance based approach, if there are now concerns about performance.

workers/ohsome_quality_analyst/geodatabase/select_shdi.sql
```diff
@@ -9,7 +9,8 @@ WITH bpoly AS (
 )
 SELECT
     SUM(ST_Area (ST_Intersection (shdi.geom, bpoly.geom)::geography) / ST_Area
-	(bpoly.geom::geography) * shdi.shdi) AS shdi
+	(bpoly.geom::geography) * shdi.shdi) AS shdi,
```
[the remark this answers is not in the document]
joker234: the `,` was added to add another parameter in the next line

workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
joker234: We agreed that we get an Exception at the moment. In the future we might handle the issue differently.

workers/ohsome_quality_analyst/indicators/building_completeness/indicator.py
joker234: @matthiasschaub please add one comment or rename `mean` to `default_value` or sth. similar. This would make the code more comprehensible :)

joker234 (approved):

workers/ohsome_quality_analyst/geodatabase/select_shdi.sql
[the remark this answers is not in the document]
Hagellach37: this was not my question: the question is about what the value is about you calculate here?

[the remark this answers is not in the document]
Hagellach37: you calculate the area of the intersection  and divide by the overall area of the hexagon and then multiply by the shdi value.

so for example: 

0.5 / 1 * 0.84 = 0.42?

but this does not make sense. The SHDI should not be manipulated like this

PR #417, opened 2022-08-28
Update api.md

ddebarros-mapaction:
### Description
On the "Request an Indicator for a custom AOI and Layer using Python and `requests` Library" example., the custom Layer call only seems to be supported for the Mapping Saturation indicator. Replaced the GhsPopComparisonBuildings indicator that was originally so this example can work directly. Also added commas that were missing on the layer definition.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -110,8 +110,8 @@ bpolys = {
     ],
 }
 layer = {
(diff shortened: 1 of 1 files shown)


matthiasschaub (approved):
Thank you for the PR!

PR #416, opened 2022-08-27
rename layer ideal_vgi_infrastructure to basic_infrastructure

Gigaszi:
### Description
Rename layer ideal_vgi_infrastructure to basic_infrastructure.

### Corresponding issue
Closes #259

### New or changed dependencies
- None

### Checklist
- [x] I have updated my branch to `main` (e.g. through `git rebase main`)
- [x] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CONTRIBUTING.md#pre-commit) before committing
- [x] I have commented my code
- [x] I have added sufficient unit and integration [tests](https://github.com/GIScience/ohsome-quality-analyst/blob/main/docs/development_setup.md#tests)
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/ohsome-quality-analyst/blob/main/CHANGELOG.md)

Please check all finished tasks. If some tasks do not apply to your PR, please cross their text out (by using `~...~`) and remove their checkboxes.

PR #426, opened 2022-09-12
fix: rename layer ideal_vgi_infrastructure

matthiasschaub:


joker234 (approved):
good catch, you *could* add `(#426)` in addition to `(#416)` in the README, but you don't have to

matthiasschaub:
> good catch, you _could_ add `(#426)` in addition to `(#416)` in the README, but you don't have to

Done

joker234 (approved):

PR #759, opened 2023-12-12
Remove black. Use ruff format instead.

matthiasschaub:


.pre-commit-config.yaml
```diff
@@ -1,14 +1,13 @@
...
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.0.275
+    rev: v0.1.7
     hooks:
+      # Run the linter.
       - id: ruff
+      # Run the formatter.
+      - id: ruff-format
+        args: ["--check"]
```
joker234, line 10: Is there a reason you don't use `--diff` as it was with `black`?
```suggestion
        args: ["--check", "--diff"]
```

joker234 (approved):

PR #765, opened 2023-12-19
feat(building-comparison): make indicator core

matthiasschaub:


ohsome_quality_api/indicators/building_comparison/metadata.yaml
```diff
@@ -4,6 +4,7 @@ building-comparison:
   projects:
     - experimental
```
matthiasschaub, line 5:
```suggestion
```

PR #847, opened 2024-11-14
build(deps): update dependencies to support Python 3.13

matthiasschaub:
* 15524ee8 - build(deps): bump ruff version
* c65a0cad - build(deps): bump asyncpg version to 0.30 
* 0226ea61 - build(deps): relax version constrain for psycopg


joker234 (approved):
had no python 3.13 to check it, but the version bumps look okay

PR #1017, opened 2026-04-28
feat: enable custom topics with own filter and title

matthiasschaub:


ohsome_quality_api/topics/presets.yaml
```diff
@@ -239,6 +256,33 @@ footpath:
     - core
     - bkg
 
+footway:
```
matthiasschaub, line 259: @Gigaszi do you know why this is in here?

PR #1018, opened 2026-05-07
fix: cycleway topic didn't ignore features that set value “no”

joker234:
### Description
The cycleway topic did include a lot of ways that are tagged with tag combinations like `highway=residential` and `cycleway:both=no`. This means it does *not* qualify as a cycleway. It just states that there isn't any cycle infrastructure. This PR proposes to change this behaviour and exclude all ways with just cycleway tags that equal to `no`.

To see the many ways this applies to and the extent of the error, one can have a look at this query. All shown ways are *not* covered by the fixed filter anymore (for Heidelberg): https://overpass-turbo.eu/s/2pma

image

### Corresponding issue
None

None

### Checklist
- [x] I have ensured my branch is mergeable with `main` (e.g. through `git rebase main` if necessary)
- [x] My code follows the [style guide](https://github.com/GIScience/ohsome-quality-api/blob/main/CONTRIBUTING.md#style-guide) and was checked with [pre-commit](https://github.com/GIScience/ohsome-quality-api/blob/main/CONTRIBUTING.md#pre-commit) before committing
- [ ] I have added sufficient unit and integration [tests](https://github.com/GIScience/ohsome-quality-api/blob/main/docs/development_setup.md#tests)
- [ ] I have added new Hurl regression tests and checked all existing [tests](https://github.com/GIScience/ohsome-quality-api/blob/main/regression-tests/README.md)
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/ohsome-quality-api/blob/main/CHANGELOG.md)


SlowMo24 (approved):
this is definitely an improvement

But as long as roads are not treated as the two-way object they are, I still feel we have not approached the scientific optimum. I.e. if a road has a bike lane on both sides it should not be counted the same as one that only has it one one side IMO. Yet, I acknowledge that that is difficult to fulfil in this context.

ohsome_quality_api/topics/presets.yaml
```diff
@@ -171,14 +171,14 @@ cycleway:
...
-    or (cycleway:left:oneway=*)
-    or (cycleway:right:oneway=*)
+    or ((cycleway:both=*) and (cycleway:both!=no))
+    or ((cycleway:right=*) and (cycleway:right!=no))
+    or ((cycleway:left=*) and (cycleway:left!=no))
+    or ((cycleway:right:lane=*) and (cycleway:right:lane!=no))
+    or ((cycleway:both:lane=*) and (cycleway:both:lane!=no))
+    or ((cycleway:left:lane=*) and (cycleway:left:lane!=no))
+    or ((cycleway:left:oneway=*) and (cycleway:left:oneway!=no))
+    or ((cycleway:right:oneway=*) and (cycleway:right:oneway!=no))
```
tyrasd: I would remove the extra set of parentheses around the `key=value` expressions:

```suggestion
    or (cycleway:both=* and cycleway:both!=no)
    or (cycleway:right=* and cycleway:right!=no)
    or (cycleway:left=* and cycleway:left!=no)
    or (cycleway:right:lane=* and cycleway:right:lane!=no)
    or (cycleway:both:lane=* and cycleway:both:lane!=no)
    or (cycleway:left:lane=* and cycleway:left:lane!=no)
    or (cycleway:left:oneway=* and cycleway:left:oneway!=no)
    or (cycleway:right:oneway=* and cycleway:right:oneway!=no)
```

technically not even the remaining parentheses are needed `… or (x and y)` is the same as `… or x and y`, but I would keep those for better clarity/readability of the filter.

tyrasd (approved):

